PR #48, opened 2018-06-13
series of changes for improved TOC

jckr:
This PR addresses a series of issues / opportunities for improved TOCs and navigation between pages. 

1) Sometimes with TOCs more than 2 level deep, the redirect on a section header isn't working. Now, instead of using a redirect, section headers feature a link to their first descendant without children. 

ex: 
title
|- subtitle
  |- child 0
  |- child 1
  |- child 2

now "title" will act as a link to "child 0", as opposed to a link to its own path which would redirect to subtitle which would redirect to child 0. 

2) Links in the TOC that would point to its own path (ie clicking on the currently open page) are replaced by plain divs to avoid unnecessary warnings. 

3) TOC that have many levels are now indented by level. 

4) Markdown pages now have a "navigation" section at the bottom to go to the previous or next page.


base/components/Toc.js
```diff
@@ -27,32 +27,49 @@ import { trees } from 'routes'
 
 const getRootPath = pathname => `/${pathname.split('/')[1]}`
 
-const renderRoute = (route, i, pathname) => (
-  <div key={i}>
+const SafeLink = ({className, name, path, pathname}) => {
```
isaacbrodsky: I'm not sure I understand what SafeLink is. It seems like it replaces the link with text if the current page is the destination?

base/components/Navigation.js
```diff
@@ -0,0 +1,41 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React from 'react'
+import { Link } from 'react-router-dom'
+
+export default function Navigation({ prev, next }) {
```
ibgreen: One line comment stating what this component is intended to do would be nice.

base/components/Toc.js
ibgreen: Agreed, a short comment before each component explaining what its purpose is could be helpful to casual readers.

base/routes.js
```diff
@@ -28,14 +28,28 @@ import jsRoutes from 'jsRoutes'
...
-  d.map(sub => {
-    const path = `${parentPath}/${sub.path || slug(sub.name, { lower: true })}`
-    return sub.children
-      ? { ...sub, path, children: generatePaths(sub.children, path) }
-      : { ...sub, path, hasToc: true }
+const generatePaths = (children, parentPath, order = []) =>
+  children.map((child, rank) => {
+    const path = `${parentPath}/${child.path || slug(child.name, { lower: true })}`
+    const newOrder = [...order, rank]
+    return child.children
```
ibgreen: Maybe a comment about the distinction that is being made here?

base/routes.js
```diff
@@ -28,14 +28,28 @@ import jsRoutes from 'jsRoutes'
...
+      ? { ...child, path, order: newOrder, children: generatePaths(child.children, path, newOrder) }
+      : { ...child, path, order: newOrder, hasToc: true }
   })
 
+const compareOrders = (route1, route2) => {
```
ibgreen: Maybe a comment to explain why we need this complicated comparison order?

base/routes.js
```diff
@@ -48,17 +62,48 @@ export const trees = mdRoutes.reduce((out, { name, path, children = [], data = [
...
 }, {})
 
-const routes = Object.keys(trees).reduce((out, key) => {
-  const { tree } = trees[key]
-  const reduced = tree.reduce((acc, cur) => acc.concat(flatten(reduction(cur))), [])
-  const final = [...reduced, reduced[0] && { path: key, redirect: reduced[0].path }]
-    .filter(d => d)
-    .sort((a, b) => (b.redirect ? -1 : a.redirect ? 1 : 0))
+const routes = Object.keys(trees)
+  .reduce((out, key) => {
```
ibgreen: Nit: I'm sure this usage of reduce is very compact but I do find it quite hard to read compared to for loops. But I suppose you are just reformatting existing code.

base/components/Toc.js
jckr: thanks. I'm going to add a comment in the code. 
The idea is that if you click on a <Link> that has a to prop equivalent to the current path, the "history" module is going to generate warnings. To avoid that I'm using the SafeLink wrapper which only creates a <Link> component for something that would take the user to a different route.

base/routes.js
jckr: lol, it's less compact that what it's replacing. but i see your point

PR #51, opened 2018-06-14
Build sitemaps

jckr:
When a user runs ocular build-docs to generate a mdRoutes.js file based on the contents of their /src/docs folder, this will now also create a sitemap.xml file based on these files. 

In other words, if the user puts their markdown files in /src/docs, ocular can not only generate a dynamic routes js file but also the corresponding sitemap.xml document so search engines can index the contents of these pages.


website/src/docs/reference/routes-and-links-in-ocular/2-documentation-routes.md
```diff
@@ -13,6 +13,7 @@ ocular build-docs
 ``` 
 
 in the CLI and this will create a mdRoutes file based on the contents of the __/src/docs/__ folder. Each markdown (.md) file will become an entry in the table of content, grouped hierarchically by folders. 
+If you have a BASEURL file in your src folder, which contains the url of the your documentation file, ocular build-docs will also generate a sitemap (__sitemap.xml__) based on these files.
```
isaacbrodsky: or are using a Github project type

Issue #52, opened 2018-07-10
Undefined variable: "$gradient-2"


javidhsueh:
I got this error in ocular@0.5.5 when starting the local server

> ERROR in ../css-loader??ref--5-1!../sass-loader/lib/loader.js??ref--5-2!../autoprefixer-loader!./base/styles/main.scss
Module build failed:
      border-image: $gradient-2;
                   ^
      Undefined variable: "$gradient-2".
      in /Users/javidhsueh/graph.gl/docs/node_modules/ocular/base/styles/_toc.scss (line 75, column 21)
 @ ./base/styles/main.scss 4:14-160 18:2-22:4 19:20-166
 @ ./base/components/App.js
 @ ./base/main.js
 @ multi ../webpack-dev-server/client?http://localhost:8080 webpack/hot/dev-server react-hot-loader/patch babel-polyfill whatwg-fetch ./base/main

ibgreen:
Related to old ocular, closing

PR #54, opened 2018-07-10
avoid scss variable if it doesnt exist

jckr:
this should fix issue #52. this condition existed at some point in the code...


jckr:
@isaacbrodsky this variable wasn't defined in earlier versions of ocular. 
The idea behind it was that instead of all repos using the same gradient and having a default, largely empty Hero component, the colors of the gradient could be generated (ie by hashing the project name) and that gradient could be used for the Hero component, so out of the box all projects won't just all look the same. I never got to redo that Hero component though and so far creating that variable has done more harm than good.

PR #56, opened 2018-08-04
Better support for jsRoutes

jckr:
Previously, jsRoutes was taken as is - there was no reprocessing at all. 
All routes that needed to appear in the table of contents had to be in mdRoutes. 

The issue there is that mdRoutes mixed 2 kind of routes - those with documentations (as intended) and those with standalone components.

With this PR, we can go back to a model where mdRoutes contains only routes with markdown files (and can easily be generated instead of manually curated)


jckr:
(nb - this will be required for https://github.com/uber/luma.gl/issues/583)

PR #65, opened 2018-08-23
better handling of empty route

jckr:
When there's a documentation route which either has neither has:
* a markdown property, which is truthy,
* a valid markdownUrl property,
* a valid component property

This results in an error. With this fix this will result in an empty page but won't crash the site, so further navigation is possible.

### DIFF
--- a/base/components/Wrapper.js
+++ b/base/components/Wrapper.js
@@ -36,8 +36,8 @@ class Wrapper extends Component {
   }
 
   render() {
-    const { markdown, markdownUrl, component } = this.props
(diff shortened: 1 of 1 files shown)


isaacbrodsky:
Could this be detected that the route isn't valid at build time and a warning emitted? Does this also cover invalid/not found routes?

jckr:
Good point. We could issue a warning if we are building routes automatically, and the file exists but is empty. If there is no file and we are building routes automatically, there will be no route so no problem. For routes in jsRoutes, which are created entirely manually, we could issue warnings in development

PR #80, opened 2019-01-17
Styles more similar to luma.gl

jckr:
This diff touches up styles a little bit to get closer to how they look in the current version of luma.gl

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9906 @@
+{
+  "name": "ocular",
+  "version": "0.6.2",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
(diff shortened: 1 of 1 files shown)


jckr:
didn't mean to add linting differences. @ibgreen  should we have a prettier or eslint config in the repo?

Pessimistress:
On the lock file: shall we use yarn to align with other frameworks?

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -55,7 +56,7 @@ const renderRoute = (route, i, pathname, depth) => {
   if (route.chapters) {
     const name = route.title;
     return (
-      <div key={i} style={{marginLeft: 10 * depth}}>
+      <div key={i} style={{ marginLeft: 10 * depth }}>
```
ibgreen: Yep, we definitely want to get prettier in place so we don't need to bike shed over spacing...

PR #81, opened 2019-01-24
Collapsible toc

jckr:
collapsible toc

This PR reproduces the collapsible TOC model of classic ocular.
If a chapter contains the page which is currently displayed it will be open, while the rest of the TOC is collapsed. 
In addition, it is possible to expand all TOC entries (or return to default mode). 

We should also have a responsive mode when this logic translates to having drop down menus with the corresponding entries.

### DIFF
--- a/modules/gatsby/src/components/layout/table-of-contents.jsx
+++ b/modules/gatsby/src/components/layout/table-of-contents.jsx
@@ -18,57 +18,118 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React, { PureComponent } from 'react'
-import cx from 'classnames'
+import React, { PureComponent } from 'react';
+import cx from 'classnames';
 import { Link } from 'gatsby';
 
-const getRootPath = pathname => `/${pathname.split('/')[1]}`
+function getRouteInfo({ route, slug }) {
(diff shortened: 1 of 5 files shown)


modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -18,57 +18,118 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React, { PureComponent } from 'react'
-import cx from 'classnames'
+import React, { PureComponent } from 'react';
+import cx from 'classnames';
```
ibgreen: Nit: `cn`? `classnames`?

modules/gatsby/src/components/layout/top-level-layout.jsx
```diff
@@ -14,6 +14,7 @@ import SEO from '../common/SEO'
 import TableOfContents from './table-of-contents'
 import ExampleTableOfContents from './example-table-of-contents'
 import Header from './header'
+import { CompositeDisposable } from 'rx';
```
ibgreen: nit: If choosing to import a component that is uncommon in our repos, maybe add a comment about what this is, to save the maintainer the burden of having to look it up.

For some reason I don't actually see it being used below...

PR #82, opened 2019-01-26
responsive header more similar to Ocular

jckr:
This is an attempt to make the header behave similarly to Ocular. 
responsive-header

In ocular headers when screen goes under a certain size there is a top right icon to open a menu that displays top navigation links. In further work we can also display table of contents in this mode (as classic Ocular does).


modules/gatsby/src/components/layout/header.jsx
```diff
@@ -53,36 +52,106 @@ import {Link} from 'react-router-dom';
...
+            ]
+          : [])
+      ]
+    };
   }
 
   renderHeader() {
     // TODO/ib - replace data with config
-    const {config = {}, pathname, isMenuOpen, opacity, stargazers_count, githubLoading} = this.props;
+    const {
```
ibgreen: You can remove the TODO

jckr: I'm going to make one more pass to include github stars

PR #83, opened 2019-01-31
rewrite local links

jckr:
This PR allows docs authors to use "physical" links (ie from markdown file to markdown file, written in a syntax compatible w/ gitHub conventions) and have them rendered by gatsby as a link from a route to another route.

### DIFF
--- a/modules/gatsby/src/gatsby-node/create-pages.js
+++ b/modules/gatsby/src/gatsby-node/create-pages.js
@@ -1,7 +1,7 @@
 const path = require('path');
 const assert = require('assert');
 
-const {log, COLOR} = require('../utils/log');
+const { log, COLOR } = require('../utils/log');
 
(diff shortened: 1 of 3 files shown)


modules/gatsby/src/gatsby-node/create-pages.js
```diff
@@ -164,12 +175,57 @@ function createDocPages({ graphql, actions }) {
         categorySet.add(edge.node.frontmatter.category);
       }
 
+      const relativeLinks = pathToSlug.reduce((prev, { source, target }) => {
```
ibgreen: Total Nit: Not a big fan of `reduce`, prefer `for of` or `forEach`. Though theoretically elegant, I find `reduce` code harder to read.

More importantly this many lines deserves to be put in their own function.

PR #84, opened 2019-02-01
Touch up markdown

jckr:
image

markdown style closer to original luma.gl / deck.gl style


modules/gatsby/src/gatsby-node/create-pages.js
```diff
@@ -119,6 +119,54 @@ function createExamplePages({ graphql, actions }) {
...
 }
 
+function addToRelativeLinks({
+  source,
+  target,
+  rootFolder,
+  edge,
+  relativeLinks
+}) {
+  // what we are doing here: for each markdown file, we create a mapping of different ways to
```
ibgreen: Looks like this contain your previous PR, maybe land and rebase?

PR #85, opened 2019-02-01
Expose bootstrap script

Pessimistress:
- Expose bootstrap script
- Run `browser-headless` test in full test mode


modules/dev-tools/package.json
```diff
@@ -16,6 +16,7 @@
     "templates"
   ],
   "bin": {
+    "ocular-bootstrap-monorepo": "./scripts/bootstrap-monorepo.sh",
```
ibgreen: I think we could just call it `ocular-bootstrap` and following your directory sniffing autodetection logic, just do nothing if not a monorepo....

PR #86, opened 2019-02-01
Support monorepo in build & clean scripts

Pessimistress:


modules/dev-tools/scripts/build.sh
```diff
@@ -1,4 +1,35 @@
...
+build_module() {
+  CONFIG=$1
+
+  BABEL_ENV=es6 npx babel src --config-file $CONFIG --out-dir dist/es6 --source-maps
+  BABEL_ENV=esm npx babel src --config-file $CONFIG --out-dir dist/esm --source-maps
+  BABEL_ENV=es5 npx babel src --config-file $CONFIG --out-dir dist/es5 --source-maps
+}
+
+if [ -d "modules" ]; then
+  # Monorepo
```
ibgreen: Nit: Would probably be a little more readable to put both cases in functions, so that

```
if [ -d "modules" ]; then
  build_monorepo()
else
  build_polyrepo()
fi
```

Hmmm: polyrepo? unirepo? ...?

PR #92, opened 2019-02-06
basic search

jckr:
ocular search 2


modules/gatsby/package.json
```diff
@@ -62,6 +62,7 @@
     "gatsby-transformer-sharp": "2.1.8",
     "inquirer": "^6.2.1",
     "lodash": "^4.17.4",
+    "lodash.debounce": "^4.0.8",
```
ibgreen: For different PR: Check if we can drop lodash and gatsby-plugin-lodash in favor or just the specific plugins. The starter I based this on used lodash but I don't think we use lodash anymore,

modules/gatsby/src/components/layout/header.jsx
```diff
@@ -99,11 +98,12 @@ export default class Header extends Component {
               <a key={link.href} href={link.href}>{link.name}</a>
             )) */
     }
-
+    // TODO - let's create the links server side, then pass them to the template as props.
     this.state = {
       links: [
         ...(hasExamples(props) ? [{ label: 'Examples', to: '/examples' }] : []),
         { label: 'Documentation', to: '/docs' },
+        { label: 'Search', to: '/search' },
```
ibgreen: Do we want this to be configurable in `ocular-config`? Or we always have search, the question is just if it is search-as-a-service or pre-generated?

modules/gatsby/src/gatsby-node/create-pages.js
```diff
@@ -248,3 +220,62 @@ function createDocPages({ graphql, actions }) {
...
+  const { createPage } = actions;
+  // TODO/ib - plugin options no longer provided when we are not a plugin
+  // We seem to be getting site metadata instead?
+  const {
+    docPages = true,
+    examplePages = true // TODO - autodetect based on DEMOS config
+  } = pluginOptions;
+
+  createIndexPage({ graphql, actions });
+  const searchPromise = graphql(`
```
ibgreen: * Put the search query and page creation code in helper function `createSearchPage`?
* Return a promise from each `create...Page` function and do a `return Promise.all([create...]);` here in the main function?

modules/gatsby/src/gatsby-node/process-nodes-json.js
```diff
@@ -25,6 +25,7 @@ module.exports.getTableOfContents = function getTableOfContents() {
 }
 
 function traverseTableOfContents(chapters, docNodes, level) {
+  debugger;
```
ibgreen: Cut

modules/gatsby/src/components/layout/header.jsx
jckr: Yes I absolutely want this to be in ocular-config. This and all the other links. I don't think we can have a systematic link to vis.gl medium blog on the header :) plus people can add links to story book, or other custom parts of website

PR #95, opened 2019-02-14
Examples homepage

jckr:
This PR addresses
- creation of an examples gallery page,
- generation of image thumbnails for the purpose of this gallery,
- fix to example runner to provide sourceLink when needed (which crashed example when external files had to be loaded)
- slght change to ocular config template to add DIR_NAME to config

### DIFF
--- a/modules/gatsby/src/gatsby-config/get-gatsby-config.js
+++ b/modules/gatsby/src/gatsby-config/get-gatsby-config.js
@@ -75,7 +75,14 @@ module.exports = function getGatsbyConfig(config) {
           path: `${__dirname}/../../static/`
         }
       },
-      // Generates gatsby nodes for files in the website's static folder
+      // Generates gatsby nodes for files in the website's src folder
+      {
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'src',
+          path: `${config.DIR_NAME}/src/`
+        }
+      },
       // Generates gatsby nodes for files in the website's static folder
       {
         resolve: 'gatsby-source-filesystem',
--- a/modules/gatsby/src/gatsby-config/get-gatsby-config.js
+++ b/modules/gatsby/src/gatsby-config/get-gatsby-config.js
@@ -79,14 +79,7 @@ module.exports = function getGatsbyConfig(config) {
           path: `${__dirname}/../../static/`
         }
       },
-      // Generates gatsby nodes for files in the website's src folder
-      {
-        resolve: 'gatsby-source-filesystem',
-        options: {
-          name: 'src',
-          path: `${config.DIR_NAME}/src/`
-        }
-      },
+      
       // Generates gatsby nodes for files in the website's static folder
       {
         resolve: 'gatsby-source-filesystem',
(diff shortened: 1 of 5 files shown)


modules/gatsby/src/gatsby-config/get-gatsby-config.js
```diff
@@ -43,10 +46,11 @@ module.exports = function getGatsbyConfig(config) {
 
       // A Gatsby plugin for styled-components with built-in server-side rendering support.
       'gatsby-plugin-styled-components',
-
+      'gatsby-plugin-sharp',
```
jckr: this is the interesting part

modules/gatsby/src/templates/examples.jsx
```diff
@@ -1,9 +1,48 @@
...
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import { Link } from 'gatsby';
 
 import { graphql } from 'gatsby';
 
-import ExampleTableOfContents from '../components/layout/example-table-of-contents'
+import ExampleTableOfContents from '../components/layout/example-table-of-contents';
+
+const colors = {
```
jckr: this is in preparation for a future where all of these variables will be in one space

modules/gatsby/src/gatsby-node/create-pages.js
```diff
@@ -67,6 +67,41 @@ function createSearchPage({ graphql, actions }) {
...
+  // with the public url of a thumbnail, resized server side by imageSharp.
+  if (
+    !allFile ||
+    !allFile.edges ||
+    !allFile.edges.length ||
+    !allImageSharp ||
+    !allImageSharp.edges ||
+    !allImageSharp.edges.length
+  ) {
+    return {};
```
ibgreen: Emit some warning?

jckr: ok. 
There is something else - if we go down have a query that says 
```
siteMetadata {
          config {
            EXAMPLES {
              image
              title
              path
            }
          }
        }
      }
```

and there is at least one example THEN that example MUST have the fields image, title and path. If there is no example with the image field, that query will result in an error. 

So we could either - 

do a try / catch pattern with this query, and if it fails, do the same query without image and without looking for thumbnails, 

or enforce having an image field in all examples.

ibgreen: Yes... for now I think it could be enough to have a minimal try catch that logs a message hinting at what went wrong.

modules/gatsby/src/gatsby-config/get-gatsby-config.js
```diff
@@ -309,6 +313,19 @@ module.exports = function getGatsbyConfig(config) {
     ]
   };
 
+  // conditional plug-ins - only added depending on options on config
+
+  if (config.DIR_NAME) {
```
jckr: w/o this check this generates an error for ocular-gatsby project that don't have a DIR_NAME specified in config. 
we can use this pattern to add plugins that won't break for older projects.

ibgreen: Do we want to warn if we don't find key variables in ocular-config?

modules/gatsby/src/gatsby-node/create-example-pages.js
```diff
@@ -0,0 +1,160 @@
...
+      component: EXAMPLE_PAGE,
+      context: {
+        slug: exampleName,
+        toc: 'examples'
+      }
+    });
+  });
+}
+
+module.exports = function prepareExamplePages({ graphql, actions }) {
```
ibgreen: Nit: Maybe dedicate this long function just to doing the query, rather than hiding an example creation stanza at the very end?

jckr: i'm not sure what's a better form. could you take a stab? 
the point of this function is to do 

query w/ images -> if success, create thumbnails, then create examples with images
|
v 
if failure, query w/o images -> if success, create examples w/o images

PR #96, opened 2019-02-14
this prevent crashes if addToRelativeLinks processes a node w/ a slug

jckr:


modules/gatsby/src/gatsby-node/create-pages.js
```diff
@@ -155,12 +155,16 @@ function addToRelativeLinks({
 
   // note that often, the physical location and the slug are the same!
   // However there is no guarantee that this will be the case.
+  
+  if (!source || !target) {
+    return {};
```
ibgreen: Can we emit a warning to help the user debug his toc/links etc?

jckr: ok

PR #98, opened 2019-02-19
Styling through styled components

jckr:
This PR lays the groundwork for having styling done (almost) entirely in CSS-in-JS as opposed to rely on SCSS. The reasoning for moving towards a CSS-in-JS solution is that we can have everything configured in a central place as opposed to have to hope that variables in the code are consistent with variables in stylesheets.

To that end, it creates a central list of styled components. These components are created, today, with styled-components but it would be easy to move to styletron if that's where we going. styled-components doesn't have to be imported anywhere else.

It also creates a theme.js file which contains primitives (descriptions of colors, spacing etc.) and a method to create a theme (one or several css properties directly usable on a component). This is copied from baseUI. 

Finally it passes that theme through the website-config context consumer. Possibly a more specialized theme consumer could be created though that's directly what baseUI does. 

Note that I haven't been able to add theme overrides to the static query in layout - any change resulting in a never ending StaticQuery(loading).

### DIFF
--- a/modules/gatsby/src/components/layout/layout.js
+++ b/modules/gatsby/src/components/layout/layout.js
@@ -4,7 +4,8 @@
 
 import React from 'react'
 import Helmet from 'react-helmet'
-import styled from 'styled-components'
+import { lightThemePrimitives, createTheme } from '../styled/theme';
+
 import {StaticQuery, graphql} from 'gatsby'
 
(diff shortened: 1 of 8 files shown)


jckr:
ok great, so we can investigate tomorrow how to make it work going forward.

Issue #101, opened 2019-03-05
thoughts on ocular TOC generation


jckr:
those are my notes as I'm researching better TOC generation for ocular. 

First, it makes no sense to have a chapter notion different from entries. they are functionally the same and they just complicate the graph. 

(What we could have is a notion of tree in the TOC. in some "legacy" ocular sites (ie vis.academy) we have several independant TOCs ie you can't navigate from a page described in one TOC to the next. those trees all have different entries in the header menu (or elsewhere). That's a P1 though)

If we use the onCreateNode callback, we can access the TOC node as it's created and directly mutate it. The problem is that the function we'd use relies on a docNodes object, which is half populated by then. 

The alternative is to work at the createPages level, where we execute a graphQl query. But then, we no longer have access to the TOC node / we can't mutate it directly. 
However, we can have the same information as a result from that query. If we can access the TOC node then great, else we'd have to create a new node and pass it the relevant information.

ibgreen:
I am not quite sure what the proposal is.

* I think that you are saying that we should allow an "entries" to recursively contain "entries" and do away with the "chapters" concept? (seems like a reasonable simplification).

* I don't think that are you proposing that we drop the `table-of-contents.json` and just work with auto discovering directory nodes?

FWIW, I will add a requirement: I'd like to be able to link in docs that are in multiple doc trees (rather than just in one `/docs` root.

The reason is that we are increasingly modularizing our repos with multiple code bases into monorepos with `modules` structure:

```
modules
  module1
     src
     test
     data
     docs???
     package.json
     README.md
```

We are collecting everything related to that module inside that folder, except the docs, and for big repos it gets hard to track and keep the docs correct when they are in deep subfolders in a different structure.

jckr:
those are just notes as I try to get the TOC generation to work. 
Yes, I think we should drop entries/chapters. it should all be entries or chapters. 
Regardless, we can't have an infinitely nested TOC. This graphql issue makes a point. https://github.com/facebook/graphql/issues/91
But we can go 3/4 levels down. It would be easier with something like 

```
entries {
   ...entriesFragment
   entries {
        ...entriesFragment
        entries {
              ...entriesFragment
              entries {
                   ...entriesFragment
              }
       }
   }
}
```
than if we have to accommodate every arrangement of chapter / entries. 

For multiple trees, I don't recommend dropping table-of-contents. I just feel the TOC could instead be:

[
   {
      tree: 'documentation', // default name
      entries: [...]
   },
[...]

PR #103, opened 2019-03-06
put back info of markdown nodes in tocnode

jckr:
This implements what's described in https://github.com/uber-web/ocular/issues/101. 

Currently, in luma.gl, table-of-contents.json doesn't match the contents of the folders. As a result there are some mismatches, for instance all the loaders docs have moved to loaders.gl. 

But besides that, the TOC has all the entries even "the ones that got away" 🎉


modules/gatsby/src/gatsby-node/process-nodes-markdown.js
```diff
@@ -105,24 +124,39 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode({ node,
...
-
-  docNodes[relPath] = node;
-}
+  if (tocNode) {
+    // this means toc node has been created. Any markdown file processed beyond this point wouldn't have its info
+    // in the toc.
+    // but we can inject it afterwards
+    const nodeToEdit = parseToc([tocNode], relPath);
+    if (nodeToEdit) {
+      nodeToEdit.childMarkdownRemark = {
```
jckr: the regular toc node generation process adds the full content of each markdown node to the toc. we don't need as much. The app will only use the title and slug of the corresponding markdown node for each toc entry.

modules/gatsby/src/gatsby-node/process-nodes-json.js
```diff
@@ -9,6 +9,7 @@ module.exports.processNewDocsJsonNode = function processNewDocsJsonNode({
 }, docNodes) {
   traverseTableOfContents(node.chapters, docNodes, 1);
   tableOfContents = node;
+  return node;
```
ibgreen: Looks like you left code after the return statement... can't imagine the linter being happy with this...

modules/gatsby/src/gatsby-node/process-nodes-markdown.js
```diff
@@ -1,37 +1,37 @@
...
   let processed = false;
 
   if (!processed) {
     switch (node.internal.mediaType) {
-    case `text/markdown`:
-    case `text/x-markdown`:
-      addSourceInstanceName(...arguments);
-      processed = true;
-      break;
+      case `text/markdown`:
```
ibgreen: I have said this before, but I think we need to get your new editor setup so it reads formatting rules from files in the repo, better avoid unrelated formatting changes. Time to implement prettier.

modules/gatsby/src/gatsby-node/process-nodes-markdown.js
```diff
@@ -43,11 +43,30 @@ module.exports.cleanupMarkdownNode = function cleanupMarkdownNode(
   //   console.warn('Ocular ignoring node',
   //     node.absolutePath, node.internal.mediaType, node.internal.type, node.sourceInstanceName);
   // }
+};
+
+function parseToc(queue, entry) {
+  // this function returns a node in the TOC that has an entry corresponding to
```
ibgreen: Nit: Comment that relates to whole function goes above the `function` keyword

modules/gatsby/src/gatsby-node/process-nodes-markdown.js
ibgreen: Suggestion: When I feel the need to add a comment like that to the PR to guide the reader, I usually realize that I would have been better of just adding it to the code so that future maintainers also benefit...

modules/gatsby/src/gatsby-node/process-nodes-json.js
jckr: ok I took a pass on that file to make it more linter friendly

modules/gatsby/src/gatsby-node/process-nodes-markdown.js
jckr: i'm all about settling for common prettier rules. we have a .eslintrc.json file in the repo but no .prettierrc or equivalent

ibgreen: In the same repo we have a module ocular-dev-tools, I believe it includes prettier in its scripts, we should update.

PR #104, opened 2019-03-06
[dev-tools] Add config system

Pessimistress:
This is a proof of concept.
Repos that use ocular-dev-tools can provide a single `ocular.config.js` at root for customizing the dev process. At run time, ocular merges these user configs with the default templates. This eliminates the need to manually create endpoints such as `babel.config.js`, `.eslintrc.js`, `test/webpack.config.js`, `test/start.js` etc.

### DIFF
--- a/modules/dev-tools/config/babel.config.js
+++ b/modules/dev-tools/config/babel.config.js
@@ -1,3 +1,5 @@
+const config = require('./ocular.config');
+
 const TARGETS = {
   chrome: '60',
   edge: '15',
@@ -7,10 +9,21 @@ const TARGETS = {
   node: '8'
 };
 
+function mergeEnvSettings(env, defaultSettings) {
+  const envConfig = config.babel[env] || {};
+  const presets = envConfig.presets || config.presets || [];
+  const plugins = envConfig.plugins || config.plugins || [];
+
+  return Object.assign({}, defaultSettings, {
+    presets: defaultSettings.presets.concat(presets),
+    plugins: defaultSettings.plugins.concat(plugins)
+  });
+}
+
 module.exports = {
   comments: false,
   env: {
-    es5: {
+    es5: mergeEnvSettings('es5', {
       presets: [
         [ '@babel/env', {
           forceAllTransforms: true,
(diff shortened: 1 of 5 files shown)


ibgreen:
> This eliminates the need to manually create endpoints such as babel.config.js, .eslintrc.js, test/webpack.config.js, test/start.js etc.

Haven't yet reviewed the details so I may be off the mark, but a top-level comment is that this could be an area where we want to carefully balance optimizing for convenience with avoiding too much "magic".

Because of this, I actually viewed the explicit need for endpoints such as babel.config.js, .eslintrc.js, test/webpack.config.js, test/start.js etc as a strength of the ocular-dev-tools design.

In my personal experience, it can be quite frustrating to start working with a repo that has a complex custom build system that someone has put together that requires me to start digging and figuring out what it does, when I just want to make some simple changes.

The presence of these files (even if they only contain a single line that reexports definitions from ocular) lets the user who is not intimately familiar with ocular see these files in the places they expect, makes it clear which tools are being used under the hood, and they see that those files just reexport configs from ocular, and they can modify those files to:

1) print&debug the values that ocular generates
2) apply their own overrides, in the standard files they are already familiar with.

Pessimistress:
I see your point about `eslintrc` and `babelrc`. I'll roll the change back. However, I feel strongly about not having `test/start.js` and `webpack.config.js` in each repo. We currently copy them all over, a lot of the modes do not work, and the same mode name lead to different results in each repo. It is extremely confusing.

ibgreen:
> . However, I feel strongly about not having test/start.js and webpack.config.js in each repo.

For sure we shouldn't duplicate the contents of those big files everywhere!

My thinking was that we would move `webpack.config.js` from test to root, and in that file, we would just have one line that reexports a webpack config imported from ocular. 

This way the user would discover all the familiar configs (eslint/babel/webpack/...) right in the root folder. No doubt about what tools are under the hood or where to start overriding.

`test/start.js` is obviously not a standard config file so we have a bit of flexibility there.

That said, I would personally find it very useful to be able to quickly add/override test "targets", and I was thinking that `test/start.js` could well follow the same pattern of having an "anchor" file that imports default behavior from ocular. 

But I am aware that you are working through a ton of issues here, if too much work to expose test/start.js right now, maybe leave that for now, it can come later when we need it.

PR #105, opened 2019-03-06
[dev-tools] Add test harness

Pessimistress:


modules/dev-tools/config/webpack.config.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
```
jckr: we'll have to run the script again once we've finished the current effort. I suspect not every file has the copyright notice anyway.

Pessimistress: Is there an official script to add these headers?

PR #107, opened 2019-03-06
[dev-tools] Fix metrics collection

Pessimistress:


modules/dev-tools/scripts/metrics.sh
```diff
@@ -53,22 +63,12 @@ print_all_loop() {
 
   print_bundle_size es5 all
   print_bundle_size esm all
-  # print_bundle_size es6 all
-
-  for FILE in test/size/*; do (
-    # [ -d $D ]
-    print_bundle_size es6 $(basename $FILE)
-  ); done
+  print_bundle_size es6 all
```
ibgreen: Does this mean that the measuring of each class in math.gl is now lost :(

Pessimistress: We can support multiple size entry points

PR #108, opened 2019-03-06
[dev-tools] Add user config for lint

Pessimistress:


modules/dev-tools/config/ocular.config.js
```diff
@@ -1,10 +1,18 @@
+const fs = require('fs');
```
ibgreen: Not sure if this a default config? Never mind, my thought was that we now have two ocular modules that have ocular config files. I think the names are just slightly different, but assuming we want to keep the files separate (I think we should), we may want to distinguish them further. `ocular-dev-tools.config.js`? and `ocular-gatsby.config.js` @jckr

modules/dev-tools/config/eslintrc.js
```diff
@@ -0,0 +1,23 @@
...
+    'prettier'
+  ],
+
+  rules: Object.assign({
+    'guard-for-in': 0,
+    'no-inline-comments': 0
+  }, lintConfig.rules),
+
+  parserOptions: Object.assign({
+    ecmaVersion: 2017
```
ibgreen: 2018 please...

PR #113, opened 2019-03-07
[dev-tools] Support babel/webpack config files in package

Pessimistress:
- babel uses `.babelrc`/`babel.config.js` in package root before falling back to default config
- webpack uses `webpack.config.js` in package root before falling back to default config
- remove babel config override from `ocular-dev-tools.config.js`
- Test script delegate the handling of unknown modes to user config files

I will add documentation in a follow up PR.

### DIFF
--- a/modules/dev-tools/config/ocular.config.js
+++ b/modules/dev-tools/config/ocular.config.js
@@ -5,7 +5,11 @@ const IS_MONOREPO = fs.existsSync(resolve('./modules'));
 
 const DEFAULT_CONFIG = {
   babel: {
-    plugins: ['version-inline']
+    configPath: getValidPath([
+      resolve('./babel.config.js'),
+      resolve('./babelrc'),
+      resolve(__dirname, './babel.config.js')
+    ])
   },
 
   lint: {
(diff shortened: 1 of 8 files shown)


modules/dev-tools/scripts/print-config.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+# Script to collect build size information
```
ibgreen: Copy paste?

modules/dev-tools/config/ocular.config.js
```diff
@@ -5,7 +5,11 @@ const IS_MONOREPO = fs.existsSync(resolve('./modules'));
 
 const DEFAULT_CONFIG = {
   babel: {
-    plugins: ['version-inline']
+    configPath: getValidPath([
+      resolve('./babel.config.js'),
+      resolve('./babelrc'),
```
ibgreen: I would drop support for this.

Pessimistress:
> If I read this correctly, this system will work without any config files in the repo, and then even when you add them, it is ocular instead of the application that controls the merging and override management.

If a config file is found in the repo, it is used **instead of** the default config. There is no merging unless the user imports and uses the default config in their own config file.

modules/dev-tools/config/ocular.config.js
Pessimistress: We still have a few repos that use `.babelrc`. It would be confusing if it's silently ignored.

PR #119, opened 2019-03-12
101 create query layer

jckr:
This is slightly less ambitious than described in #101. 
This creates a separate component, QueryLayer. By the same token as the previous solution, it has to coexist in ocular-gatsby and in the local tree. It can be copied during the initial setup though. 
I haven't changed the syntax of the queries yet - one step at a time. But now there's just one place where this data has to be edited. I've also added the possibility to get theme overrides from the config. 

Luma.gl should get a matching update to showcase this.


modules/gatsby/src/components/query-layer.jsx
```diff
@@ -0,0 +1,99 @@
+// NOTE: Dummy component, just to get gatsby to compile the StaticQuery
+
```
ibgreen: Add more comment, here explaining that this must be identical to the query in ocular-gatsby and must not be modified.

PR #127, opened 2019-03-23
Use theme fallbacks

jckr:
with styled components not using theme from props the gatsby builds pass. from there we can debug and evaluate the best way forward.


jckr:
yeah. from there we can debug what props do these components have access to in the production build and specifically why can't they see the theme.

PR #131, opened 2019-03-28
Schema customization

jckr:
This implements schema customization for the table of content query.
the schema is no longer inferred from the contents of the file but can be defined beforehand.


modules/gatsby/src/gatsby-node/index.js
```diff
@@ -2,54 +2,115 @@ const {log, COLOR} = require('../utils/log');
...
+  const {name} = type;
+  const {createNodeField} = actions;
+  if (name === 'MarkdownRemark') {
     addSiblingNodes(createNodeField);
   }
-};
+}
+
+function sourceNodes({actions}) {
+  const {createTypes} = actions;
```
ibgreen: I recommend keeping this file small and focused on defining the gatsby callbacks, and delegating to functions in other files for actual processing.

I.e. create a new file `define-graphql-schema.js` or similar and move the implementation there.

modules/gatsby/src/gatsby-node/index.js
```diff
@@ -2,54 +2,115 @@ const {log, COLOR} = require('../utils/log');
...
+      chapters: [lvl1Chapter]
+      entries: [Entry]
+    }
+
+    type DocsJson implements Node {
+      chapters: [lvl1Chapter]
+      entries: [Entry]
+    }
+    `;
+  log.log({color: COLOR.RED}, `Created new typedefs`, typeDefs)();
```
ibgreen: I would use RED for errors, YELLOW for warnings, in this case I would either use more neutral color, or none. 

Maybe change the message to say 'Set up graphql schemas`?

modules/gatsby/src/gatsby-node/index.js
```diff
@@ -2,54 +2,115 @@ const {log, COLOR} = require('../utils/log');
...
+function setFieldsOnGraphQLNodeType({type, actions}) {
+  const {name} = type;
+  const {createNodeField} = actions;
+  if (name === 'MarkdownRemark') {
     addSiblingNodes(createNodeField);
   }
-};
+}
+
+function sourceNodes({actions}) {
```
ibgreen: Let's put the brief comment above each callback stating when gatsby calls this particular callback.

modules/gatsby/src/gatsby-node/index.js
```diff
@@ -2,54 +2,115 @@ const {log, COLOR} = require('../utils/log');
...
+      chapters: [lvl1Chapter]
+      entries: [Entry]
+    }
+    `;
+  log.log({color: COLOR.RED}, `Created new typedefs`, typeDefs)();
+  createTypes(typeDefs);
+}
 
 const GATSBY_NODE_CALLBACKS = {
   onCreateWebpackConfig,
```
ibgreen: Unrelated peeve: the webpack config is logged twice, maybe add a boolean flag to that callback and log only once?

modules/gatsby/website-template/src/components/site-query.jsx
```diff
@@ -0,0 +1,102 @@
+// This component contains the StaticQuery needed to provide data for the layout components.
```
ibgreen: We might want to have ocular init copy this one from the source tree rather than duplicate it into templates?

jckr:
I've updated the PR w/ schema for siteMetadata / config. 

Note that the DOCS property is not used. gatsby reads documentation files (markdown and json) from the DOCS_FOLDER and parse 'table-of-contents.json' file from that folder.

PR #133, opened 2019-03-29
publish: force publish all monorepo packages with exact dependencies

ibgreen:
While trying to stabilize deck.gl, I have found myself completely unable to control luma.gl dependencies. 

I cannot go back to a stable/predictable version of deck.gl even if I pin down all luma dependencies in deck as yarn will still pull in later versions, and the mismatched version names make it very hard to see at a glance if whatever got installed can be expected to work. 

Hopefully this change will provide the necessary control.

### DIFF
--- a/modules/dev-tools/CHANGELOG.md
+++ b/modules/dev-tools/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG (ocular-dev-tools)
 
+## v0.0.17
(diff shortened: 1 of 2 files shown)


georgios-uber:
Can we also have an option to override `--cd-version`?

ibgreen:
> Can we also have an option to override --cd-version?

PRs are welcome

PR #136, opened 2019-03-30
less aggressive maxwidth

jckr:
now for a page with a table of content, the total max width of TOC + content is 1280px. previously, the max width of the TOC and content was set individually and content was limited to 600px.


ibgreen:
@jckr Sorry, I pressed the wrong button, was going to approve but ended up merging directly... I suspect that subconsciously I was eager to get this landed!

PR #138, opened 2019-04-09
disable passing theme through props to make builds pass

jckr:


jckr:
this is doing what #127 was doing.

PR #143, opened 2019-04-15
Configurable example template

### LINKED ISSUE #125: ocular-gatsby: Overriding components fails during static build
These are issues that arise when building a website (ie yarn build) vs running the development version.

```
0 | function getReactComponent(name, defaultComponent) {
  11 |   const component = registry[name] || defaultComponent;
> 12 |   assert(component !== undefined, name);
     |   ^
  13 |   return component;
  14 | }
  15 |


  WebpackError: AssertionError [ERR_ASSERTION]: EXAMPLES

  - component-registry.js:12 getReactComponent
    lib/Users/jerome/code/ocular/modules/gatsby/src/utils/component-registry.js:12:3

  - component-registry.js:18 getHeroExample
    lib/Users/jerome/code/ocular/modules/gatsby/src/utils/component-registry.js:18:20

  - hero.jsx:10 Hero.renderPage
    lib/Users/jerome/code/ocular/modules/gatsby/src/components/hero.jsx:10:5

  - hero.jsx:34 Object.children
    lib/Users/jerome/code/ocular/modules/gatsby/src/components/hero.jsx:34:26

```

in the getHeroExample function of component-registry we call 
```
const EXAMPLES = getReactComponent('EXAMPLES');
```
This may result in an error if EXAMPLES is not found in the registry. We can avoid this by adding a default argument like so: 
```
const EXAMPLES = getReactComponent('EXAMPLES', {});
```
but maybe we can have a safer default Hero component that doesn't try to load an example at all, and let that feature to be only overridable from the local tree.

Once this is fixed - 

```
 121 | // search
  122 |
> 123 | export const SearchContainer = styled.div(props => ({
      |                                                     ^
  124 |   position: 'relative',
  125 |   height: props.theme.sizing.scale1000,
  126 |   marginBottom: '20px',


  WebpackError: TypeError: Cannot read property 'scale1000' of undefined

  - index.js:123
    lib/Users/jerome/code/ocular/modules/gatsby/src/components/styled/index.js:123:53
```

In the styled components file, somehow the theme doesn't work. I don't know how / why this is different between the dev and prod version. 

One thing that I've tried is go all in and replace styled components with baseUI using https://baseweb.design/getting-started/installation/ recommendations for Gatsby. 

This makes for a simpler code btw, because we no longer have to worry about passing the theme everywhere. As long as we use 
```
``` 

all the styled components have a $theme prop which is pre-populated with the BaseUI theme. 
This works fine on the development side. 

however with build, I get: 

```
 10 | function driver(style, styletron) {
  11 |   const tx = renderDeclarativeRules(style, styletron);
> 12 |   return styletron.renderStyle(tx);
     | ^
  13 | }
  14 | function getInitialStyle() {
  15 |   return {};


  WebpackError: TypeError: Cannot read property 'renderStyle' of undefined

  - index.es.js:12 driver
    [gatsby]/[styletron-standard]/dist/index.es.js:12:1

  - index.es.js:278 Object.StyledElement.Object [as children]
    [gatsby]/[styletron-react]/dist/index.es.js:278:1


  - bootstrap:24 a.read
    lib/webpack/bootstrap:24:1

  - bootstrap:36 renderToString
    lib/webpack/bootstrap:36:1
```

ibgreen:
# Fixes #125  

The component registry we use to configure website does not work during static page generation. 

Being able to redefine the page templates seems like a viable alternative route.

If we agree on this approach, we should probably make all templates overridable.

### DIFF
--- a/modules/gatsby/src/components/layout/header.jsx
+++ b/modules/gatsby/src/components/layout/header.jsx
@@ -54,7 +54,7 @@ import {Link} from 'react-router-dom';
 function hasExamples(props) {
   const {config = {}} = props;
   const {EXAMPLES} = config;
-  return !(EXAMPLES.length === 0 || EXAMPLES[0].title === 'none');
+  return EXAMPLES.length > 0;
 }
 
 function HeaderLink({to, href, label, classnames}) {
--- a/modules/gatsby/src/gatsby-node/create-example-pages.js
+++ b/modules/gatsby/src/gatsby-node/create-example-pages.js
@@ -40,8 +40,8 @@ function getExampleThumbnails({ allFile, allImageSharp }) {
   return pathLookup;
 }
 
-function createExamplePages(EXAMPLES, thumbnailsPublicUrls = {}, createPage) {
-  if (EXAMPLES.length === 0 || EXAMPLES[0].title === 'none') {
+function createExampleGalleryPage(EXAMPLES, thumbnailsPublicUrls = {}, createPage) {
+  if (EXAMPLES.length === 0) {
     return;
   }
   // matches public urls to paths of images
@@ -58,7 +58,9 @@ function createExamplePages(EXAMPLES, thumbnailsPublicUrls = {}, createPage) {
       examples: examplesWithImage
     }
   });
+}
 
+function createIndividualExamplePages(EXAMPLES, createPage) {
   EXAMPLES.forEach(example => {
     const exampleName = example.title;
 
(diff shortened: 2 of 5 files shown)


modules/gatsby/src/gatsby-node/create-example-pages.js
```diff
@@ -67,9 +69,12 @@ function createExamplePages(EXAMPLES, thumbnailsPublicUrls = {}, createPage) {
       `Creating example page ${JSON.stringify(example)}`
     )();
 
+    const {ocularConfig} = global;
+    const exampleComponentUrl = (ocularConfig && ocularConfig.EXAMPLE_TEMPLATE_URL) || EXAMPLE_PAGE;
```
ibgreen: This is the interesting part

modules/gatsby/src/components/layout/header.jsx
```diff
@@ -54,7 +54,7 @@ import {Link} from 'react-router-dom';
 function hasExamples(props) {
   const {config = {}} = props;
   const {EXAMPLES} = config;
-  return !(EXAMPLES.length === 0 || EXAMPLES[0].title === 'none');
+  return EXAMPLES.length > 0;
```
jckr: feels good to get rid of this

PR #158, opened 2019-05-09
Fix module aliasing for probe.gl

Pessimistress:
Move `moduleAlias.addAliases` to after all `require` is done.

This was breaking probe.gl tests because `@probe.gl/test-utils` is aliased.

### DIFF
--- a/modules/dev-tools/node/test.js
+++ b/modules/dev-tools/node/test.js
@@ -3,13 +3,6 @@
 // Enables ES2015 import/export in Node.js
 const {resolve} = require('path');
 
-/* global process */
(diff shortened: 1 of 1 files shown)


ibgreen:
Probably would have been good to add a comment to maintainers explaining that this order is important. 

Also shouldn't things still work even if the test-utils are aliased? If a subtle issue might also have been worth mentioning in a comment.

Pessimistress:
When test-utils is aliased to the source directory, node fails on the usage of `import`. This script cannot import reify because it may conflict with @babel/register if the user repo chooses so.

PR #161, opened 2019-05-09
Support standard BABEL_ENV="development" usage

twojtasz:


Pessimistress:
> my biggest problem with ocular is that it compiles code in development mode

This is not true. The use case that this change is trying to address uses `@bable/register`, and that is how babel config is accessed in dev mode. Unless that module is imported, no transpilation is done on your source.

ibgreen:
> This is not true. The use case that this change is trying to address uses @bable/register, and that is how babel config is accessed in dev mode. Unless that module is imported, no transpilation is done on your source.

> You may also manually set the BABEL_ENV environment variable when running your test command to control which config to use.

Thanks for the tips. I had assumed the addition of `babel/register` in loaders.gl was required by `ocular-dev-tools`, happy to hear that it is not.

ibgreen:
Added your comments to docs in https://github.com/uber-web/ocular/pull/162

PR #164, opened 2019-05-11
replacing styled-components by styletron/baseweb

jckr:
(this is not extensively tested and not ready to land)

This is an attempt to use styletron / baseweb as the main way styles are expressed. if done right this will result in:
- one single source of truth for styling - the theme,
- more concise code,
- it goes to show the craft of uber front end tech


modules/gatsby/package.json
```diff
@@ -72,16 +73,21 @@
     "prismjs": "^1.15.0",
     "probe.gl": "^3.0.1",
     "prop-types": "^15.6.2",
+    "react": "^16.8.6",
```
ibgreen: These were previously moved to `peerDependencies` to let the application control React versions.

modules/gatsby/src/templates/doc-n.jsx
```diff
@@ -20,39 +34,60 @@ export const query = graphql`
   }
 `;
 
-function replaceLinks(props) {
-  const { html } = props.data.docBySlug;
-  const { relativeLinks } = props.pageContext;
+// function replaceLinks(props) {
```
ibgreen: Add comment or cut

jckr:
image

modules/gatsby/package.json
jckr: reverted to that

modules/gatsby/src/templates/doc-n.jsx
jckr: removed

modules/gatsby/src/components/layout/layout.jsx
```diff
@@ -3,7 +3,13 @@
 // each page using the gatsby browser/SSR `wrapPage` callback.
 
 import React from 'react';
+<<<<<<< HEAD
 import {lightThemePrimitives, createTheme} from '../styled/theme';
+=======
+import {createTheme} from 'baseui';
+import {primitives} from '../styled/theme';
+
+>>>>>>> use baseui typography for markdown, replace clan pro by uber move
```
javidhsueh: Resolve conflicts?

gergelyke:
🤗

ibgreen:
@gergelyke 	Unfortunately we had to back out the styletron support due to issues during gatsby build. Perhaps we could team up and make another attempt to get this landed?

gergelyke:
that would be great - what issues did you run into?

PR #170, opened 2019-05-15
Fix site query to get the additional links

javidhsueh:


modules/gatsby/src/components/layout/header.jsx
```diff
@@ -121,7 +121,7 @@ export default class Header extends Component {
               }
             ]
           : []),
-        ...(props.config && props.config.ADDITIONAL_LINKS)
+        ...(props.config && props.config.ADDITIONAL_LINKS.map(link => ({...link, label: link.name})))
```
ibgreen: It gets a bit hard to read with all these nested expressions. Maybe create each sublist in advance and assign to a var and just do ```, ...GITHUB_LINKS, ...ADDITIONAL_LINKS]```.

Also since this is getting long, recommend breaking out a function `generateHeaderLinks()`.

modules/gatsby/src/components/layout/header.jsx
```diff
@@ -72,57 +81,48 @@ function HeaderLink({to, href, label, classnames = ''}) {
   );
 }
 
+/**
+ * Generate all the links in the header.
+ * @param  {Object} props Input props which includes site config.
+ * @return {Array}  Array of link object ({label, to, href, classnames})
+ */
+function generateHeaderLinks(props) {
```
javidhsueh: @ibgreen I've addressed your comment.

PR #174, opened 2019-05-16
Revert `sharp` module to v0.22.0 to support jpeg images

javidhsueh:
`sharp` v0.22.1 causes a regression on supporting jpeg images. Reverting `sharp` back to v0.22.0 to avoid regression.

### DIFF
--- a/modules/gatsby/package.json
+++ b/modules/gatsby/package.json
@@ -33,7 +33,7 @@
     "gatsby": "^2.3.0",
     "react": "^16.6.0",
     "react-dom": "^16.6.0",
-    "sharp": "^0.22"
+    "sharp": "^0.22.0"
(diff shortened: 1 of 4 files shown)


modules/gatsby/package.json
```diff
@@ -33,7 +33,7 @@
     "gatsby": "^2.3.0",
     "react": "^16.6.0",
     "react-dom": "^16.6.0",
-    "sharp": "^0.22"
+    "sharp": "^0.22.0"
```
ibgreen: Not sure if this makes much of a difference... Should we force it to 0.22.0 for now to make sure apps update their deps?

javidhsueh: done.

PR #176, opened 2019-05-18
Only install one copy of chromium

Pessimistress:
Requires https://github.com/uber-web/probe.gl/pull/74


Pessimistress:
puppeteer is a dependency of @probe.gl/test-utils. In the short term, there is no way for us to drop it without breaking semver.

You can set the `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` environment variable globally on your box to avoid this issue.

PR #177, opened 2019-05-19
[gatsby] add github stars & contributors, fix blog url

georgios-uber:


modules/gatsby/src/components/github/github-contributors.jsx
```diff
@@ -0,0 +1,43 @@
...
+export default class GithubContributors extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      response: window.fetchContributorsResponse
+    };
+  }
+
+  componentDidMount() {
+    if (window.fetchContributorsResponse) return;
```
ibgreen: There should be curly braces here. Looks like lint is completely broken.

modules/gatsby/src/components/github/github-contributors.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+    const {project} = this.props;
+    fetch(`https://api.github.com/repos/${project}/contributors`)
+      .then(response => response.json())
+      .then(response => {
+        window.fetchContributorsResponse = response;
```
ibgreen: Why is this a global variable? Is this class being mounted and unmounted? Add a comment?

Can we declare a "local" global variable inside this file (or store the response on the `GithubContributors` class as static field rather than referencing `window` directly?

Maybe add a utility file/functions to query and cache github data?

modules/gatsby/src/components/github/github-contributors.jsx
```diff
@@ -0,0 +1,43 @@
...
+          rel="noopener noreferrer"
+          href={contributor.html_url}
+          className="Contributor"
+          key={contributor.id}
+        >
+          <img src={contributor.avatar_url} width="100%" alt="" />
+          <span>{contributor.login}</span>
+        </a>
+      ) : (
+        <div className="Contributor" key={contributor.id} />
```
ibgreen: How does this work? `contributor` is "falsy" in this branch. Won't `contributor.id` crash?

modules/gatsby/src/components/home.jsx
```diff
@@ -21,81 +21,59 @@
...
-  renderPage({config}) {
-    const {contributors = []} = this.props;
-
-    const {
-      HOME_HEADING,
-      HOME_RIGHT = '...',
-      HOME_BULLETS = [],
-      PROJECT_TYPE
-    } = config;
+function renderPage({config}) {
```
ibgreen: Nits:
- When using functional components we still give them class names instead of function name, i.e nouns instead of verbs. E.g. `HomePage` instead of `renderPage`. 
- We do this because to the user it is an implementation detail how this component was implemented, not something we want to highlight by using different naming conventions.

modules/gatsby/src/components/layout/header.jsx
```diff
@@ -126,6 +128,15 @@ export default class Header extends Component {
     };
   }
 
+  renderStars() {
+    const {config} = this.props;
+    if (config.PROJECT_TYPE === 'github') {
```
ibgreen: Nit: We could either
- let the `<Github...>` react components know how to read the config (current design), and then we can move the `if (config.PROJECT_TYPE === 'github')` statement into the component.
- Alternative make the github components unaware of the config and destructure it and supply specific params before we call them.
- Mixing like this does not seem ideal.

modules/gatsby/src/components/github/github-contributors.jsx
georgios-uber: Copy-pasted this.

modules/gatsby/src/components/home.jsx
georgios-uber: This is not a functional component.

modules/gatsby/src/components/layout/header.jsx
georgios-uber: Maybe refactor later.

tgorkin:
Thanks for fixing the blog URL!  Do you have a screenshot of what these changes look like?

georgios-uber:
@tgorkin check https://nebula.gl/ - I updated the website to gatsby (might need to do hard refresh)

PR #180, opened 2019-05-20
Update website template in ocular-gatsby folder

javidhsueh:
- Update template config
 - Upgrade dependencies.

### DIFF
--- a/modules/gatsby/website-template/ocular-config-template.js
+++ b/modules/gatsby/website-template/ocular-config-template.js
@@ -9,11 +9,19 @@
 
 module.exports = init => `
 module.exports = {
+  // Adjusts amount of debug information from ocular-gatsby
+  logLevel: 4,
+
   DOC_FOLDER: '${init.websiteFolder}/../docs/',
   ROOT_FOLDER: '${init.websiteFolder}/../',
 
   EXAMPLES: [
-    // {title: 'my example', path: 'examples/my-example/', image: 'images/my-example.jpg'}
+    // {
+    //   title: 'my example',
+    //   path: 'examples/my-example/',
+    //   image: 'images/my-example.jpg',
+    //   componentUrl: '../examples/app.js'
+    // }
   ],
   DOCS: [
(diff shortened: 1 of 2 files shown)


modules/gatsby/website-template/ocular-config-template.js
```diff
@@ -34,6 +42,7 @@ module.exports = {
   PROJECT_DESC: '${init.desc ||
     'A documentation website for Ocular, made with Ocular'}',
   WEBSITE_PATH: '${init.path || '/website/'}',
```
javidhsueh: Is WEBSITE_PATH deprecated? I don't see it in the documentation.

modules/gatsby/website-template/ocular-config-template.js
```diff
@@ -9,11 +9,19 @@
...
   EXAMPLES: [
-    // {title: 'my example', path: 'examples/my-example/', image: 'images/my-example.jpg'}
+    // {
+    //   title: 'my example',
+    //   path: 'examples/my-example/',
+    //   image: 'images/my-example.jpg',
+    //   componentUrl: '../examples/app.js'
+    // }
   ],
   DOCS: [
```
javidhsueh: Maybe we should also provide a sample table-of-contents.json?

modules/gatsby/website-template/ocular-config-template.js
ibgreen: > Is WEBSITE_PATH deprecated? I don't see it in the documentation.

I think it is no longer used. We could remove it and issue a warning when we first load ocular-config.

We should probably have an object with default values for all supported keys and emit warnings if ocular-config contains any unknown keys. That could save users a lot of time. I will open an issue on that.

modules/gatsby/website-template/ocular-config-template.js
ibgreen: Yes it could for instance just contain the top level chapters. Not sure if that would build though...

PR #182, opened 2019-05-20
better design for github-contributors, add mapbox key env

georgios-uber:
allow additional link positions, fix homepage...

### DIFF
--- a/modules/gatsby/src/components/github/github-contributors.jsx
+++ b/modules/gatsby/src/components/github/github-contributors.jsx
@@ -1,8 +1,37 @@
 import React, {Component} from 'react';
+import styled from 'styled-components';
 
 // Github api has rate-limits. We want to cache the response
 // as much as we can. This component gets re-mounted multiple times.
 let cachedResponse = null;
+
+const ContribLink = styled.a`
+  margin: 10px;
+`;
+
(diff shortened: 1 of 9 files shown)


modules/gatsby/src/components/github/github-contributors.jsx
```diff
@@ -28,19 +57,20 @@ export default class GithubContributors extends Component {
 
   render() {
     const {response} = this.state;
```
ibgreen: Just having `response` in state looks a bit unfinished. Can't the part of code that loads do the data verification and store `contributors` in state?

modules/gatsby/src/components/site-query.jsx
```diff
@@ -41,6 +41,7 @@ const QUERY = graphql`
         ADDITIONAL_LINKS {
           name
           href
+          index
```
ibgreen: If you are adding features, please document them!

georgios-uber: Done

modules/gatsby/src/components/github/github-contributors.jsx
georgios-uber: Probably in another diff.

PR #186, opened 2019-05-22
Run prettier on test folder

javidhsueh:


modules/gatsby/test/utils/example-registry.spec.js
```diff
@@ -2,16 +2,28 @@
 import test from 'tape-catch';
 
 import {
-	getHeroExample,
-	registerReactComponent,
-	registerDefaultReactComponent,
-	getReactComponent
+  getHeroExample,
```
ibgreen: These methods are obsolete since this configuration mechanism didn't work in static builds - we are using the `componentUrl` to customize examples. We should completely remove these methods.

javidhsueh: So you're saying we should remove this test now? @ibgreen

PR #188, opened 2019-05-22
Support both locally and globally installed Chromium

Pessimistress:
Use locally installed version if available, otherwise use the global copy.


javidhsueh:
Maybe we should update the README to instruct users how to install puppeteer?

Pessimistress:
> Maybe we should update the README to instruct users how to install puppeteer?

puppeteer comes with `@probe.gl/test-utils` which is a peer dependency.

PR #189, opened 2019-05-22
 Improve the documentation for configuring tests

javidhsueh:


docs/dev-tools/configuring-tests.md
```diff
@@ -14,8 +14,18 @@ While Node 12 will soon enable `import`/`export` by default, you will typically
...
-* import `reify` - This makes Node.js understand `import`/`export`, but otherwise does not transpile your code. This is a great option if you want to test your source code directly, either because you want to debug untranspiled code, or you want to ensure that your code runs untranspiled to ensure you don't use unsupported syntax.
-* import `@babel/register` - This option runs the babel transpiler. This is ideal if you want to use non-standard syntax such as stage-x babel plugins, flow etc.
+* `require('reify')` - This makes Node.js understand `import`/`export`, but otherwise does not transpile your code. This is a great option if you want to test your source code directly, either because you want to debug untranspiled code, or you want to ensure that your code runs untranspiled to ensure you don't use unsupported syntax. You can simply require `reify` at the entry point of your test:
+
+```
+// test/index.js
+require('reify');
+
+// start to require your tests
+require('./test1.js');
```
ibgreen: Note: I believe that you cannot use `import` in the same file, only inside files `required` after requiring `reify`.

There is another similar module `esm` that looks more "official" but I never got it to work.

javidhsueh: You're right, you can't use `import` in the same file. That's why I use `require` for importing tests here

ibgreen: The reason I mentioned it was that I thought it might be worth adding that comment to the docs ;)

PR #190, opened 2019-05-23
Miscellaneous fixes for luma.gl issues

tgorkin:
- Fix github URL in header
- Fix duplicate key react warning in example gallery
- Add console warning on missing entries in table of contents.

### DIFF
--- a/modules/gatsby/src/components/layout/table-of-contents.jsx
+++ b/modules/gatsby/src/components/layout/table-of-contents.jsx
@@ -183,19 +183,21 @@ const renderRoute = ({route, index, depth, slug, fullyExpanded}) => {
           style={{maxHeight: getHeight(route)}}
         >
           <ul>
-            {route.entries.map((r, idx) => {
-              if (!r.childMarkdownRemark) {
-                console.warn(`Missing content for entry ${idx} in chapter ${route.title}`, route);
+            {route.entries.map((childRoute, idx) => {
+              if (!childRoute.childMarkdownRemark) {
+                console.warn(
+                  `Missing content for entry ${idx} in chapter ${route.title}`,
+                  route
+                );
               }
               return renderRoute({
-                route: r,
+                route: childRoute,
                 fullyExpanded,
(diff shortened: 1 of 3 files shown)


modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -183,14 +183,18 @@ const renderRoute = ({route, index, depth, slug, fullyExpanded}) => {
           style={{maxHeight: getHeight(route)}}
         >
           <ul>
-            {route.entries.map((r, idx) =>
-              renderRoute({
+            {route.entries.map((r, idx) => {
```
ibgreen: Nit: Not your code, but I am pretty OCD about typing out variables. `route, index`. This is likely from the old ocular code that never went through code review.

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -183,14 +183,18 @@ const renderRoute = ({route, index, depth, slug, fullyExpanded}) => {
...
+              if (!r.childMarkdownRemark) {
+                console.warn(`Missing content for entry ${idx} in chapter ${route.title}`, route);
```
ibgreen: Nit: We use a probe.gl `log` instance in most places. Search for `log.warn`

tgorkin: I saw the use of probe's log.warn in logic that I believe is run at "build time" (see process-nodes-json.js), but I see console.warn being used in "run time" components (see example-n.jsx).  Just want to double check that there isn't an explicit decision to keep these logging concepts separate, perhaps to avoid pulling in probe dependencies at runtime?

ibgreen: @tgorkin Yes, I see your point. There is a case for having run-time logs. 

Static HTML generation does blur the lines between runtime and build time, but using console inside the React component is probably the right choice..

PR #195, opened 2019-05-26
Enforce that a ImageSharp type exists in the schema

javidhsueh:
Per discussion on gatsby repo:
https://github.com/gatsbyjs/gatsby/issues/14304


modules/gatsby/src/gatsby-config/get-gatsby-config.js
```diff
@@ -339,6 +339,16 @@ module.exports = function getGatsbyConfig(config) {
         path: `${paddedConfig.DIR_NAME}/static/`
       }
     });
+    gatsbyConfig.plugins.push({
+      resolve: 'gatsby-source-filesystem',
+      options: {
+        name: 'images',
+        path: `${paddedConfig.DIR_NAME}/static/images`,
```
ibgreen: - Should we `fs.stat` these directories before we add the plugins?
- Or `fs.stat` and warn if they don't exist?
- Or `fs.stat` and create them if they don't exist?

modules/gatsby/src/gatsby-node/source-nodes.js
```diff
@@ -117,6 +117,10 @@ function sourceNodes({actions}) {
       chapters: [lvl1Chapter]
       entries: [Entry]
     }
+
+    type ImageSharp implements Node {
+      id: ID!
```
ibgreen: What does the exclamation markd do? Only used for this type... 

This is a `.js` file, so I assume you could add some comments above the string and let others know.

javidhsueh: The exclamation mark means the id can be empty.

PR #202, opened 2019-06-04
Add post-install script to copy site-query.jsx

javidhsueh:
'install' will be called for the install and post-install stages of the lifecycle
https://docs.npmjs.com/misc/scripts

### DIFF
--- a/modules/gatsby/bin/install.js
+++ b/modules/gatsby/bin/install.js
@@ -1,0 +1,38 @@
+#!/usr/bin/env node
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


modules/gatsby/package.json
```diff
@@ -27,7 +27,9 @@
     "publish-beta": "yarn test && npm publish --tag beta",
     "publish-prod": "yarn test && npm publish",
     "lint": "eslint src",
-    "test": "node test/index.js"
+    "test": "node test/index.js",
+    "-------- NPM LIFECYCLE --------": "",
```
georgios-uber: make this a comment

modules/gatsby/bin/install.js
```diff
@@ -0,0 +1,38 @@
...
+// THE SOFTWARE.
+
+const { readFileSync, writeFileSync } = require('fs')
+
+const SOURCE_DIR = `${__dirname}/../`;
+const DIR_PATH = process.env.PWD;
+
+// copy required files for migration
+const FILENAMES = [
+  'src/components/site-query.jsx'
```
georgios-uber: let's also install gatsby-config gatsby-browser gatsby-ssr

georgios-uber: These have to reference the ones inside ocular-gatsby

modules/gatsby/bin/install.js
```diff
@@ -0,0 +1,41 @@
...
+
+const { readFileSync, writeFileSync } = require('fs')
+
+const SOURCE_DIR = `${__dirname}/..`;
+const DIR_PATH = process.env.PWD;
+
+// copy required files for migration
+const FILENAMES = [
+  {dir: `${SOURCE_DIR}/website-template`, filename: 'gatsby-browser.js'},
+  {dir: `${SOURCE_DIR}/website-template`, filename: 'gatsby-config.js'},
```
ibgreen: in contrast to `site-query` which is just a work-around hack, the `gatsby-*` files are intended to be overridable by the user. In particular `gatsby-config.js` where the user would add additional plugins. It feels strange to just clobber these on every upgrade...

georgios-uber: gatsby-config.js should be like this:
```
// WARNING: DO NOT MODIFY THIS FILE MANUALLY IT WILL BE OVERWRITTEN.
const ocularConfig = require('...')
const userConfig = require('gatsby-config-user.js')
module.exports = lodash.merge(ocularConfig, userConfig)
```

georgios-uber: same for gatsby-ssr and others.

PR #203, opened 2019-06-04
Rename post-install script, add warnings in the files will be overwritten

javidhsueh:
Per our discussion in #202


javidhsueh:
Or maybe we can just override the site-query.jsx for now? Before we figure out how to import it from ocular-gatsby, people still need to copy over the site-query.jsx to their folder. This post-install script can at least reduce some friction when migrating to a new version of ocular-gatsby

ibgreen:
> Or maybe we can just copy the site-query.jsx for now?

Yes. I am also fine with copying the others IF they don't already exist.

BTW - the version of `gatsby-ssr` in ocular looks different from the one the @georgios-uber got to work. Might be worth making sure they are identical?

javidhsueh:
> > Or maybe we can just copy the site-query.jsx for now?
> 
> Yes. I am also fine with copying the others IF they don't already exist.
> 
> BTW - the version of `gatsby-ssr` in ocular looks different from the one the @georgios-uber got to work. Might be worth making sure they are identical?

The problem is that GraphQL will error out silently without any warning due to the query we give. It happens very often that I need to copy the `site-query.jsx` from ocular repo to override my local copy. 

My point is that the post-install script should override the `site-query.jsx` no matter it exists or not.
@ibgreen

javidhsueh:
Another question I have is that, do we need to copy `gatsby-node.js`?

ibgreen:
> My point is that the post-install script should override the site-query.jsx no matter it exists or not.

Yes 100% agree. In case of `site-query.jsx`, it is just our hack and not part of the gatsby configuration file set. Ideally it should be `.gitignored` by the user and just copied in freshly on each install.

My point was that the other files are different, they are the documented gatsby entry points, and the user may want to edit them. But that comment does not apply to `site-query.jsx`.

Thanks for double checking!

modules/gatsby/bin/post-install.js
```diff
@@ -19,23 +19,35 @@
...
-  console.log('Migrating', `${DIR_PATH}/${f.filename}`)
-  writeFileSync(`${DIR_PATH}/${f.filename}`, file);
+  const targetPath = `${DIR_PATH}/${f.filename}`;
+  // copy the file if it doesn't exist
+  if (!existsSync(targetPath)) {
+    const sourcePath = `${f.dir}/${f.filename}`;
+    const file = readFileSync(`${f.dir}/${f.filename}`);
+    console.log('Migrating', targetPath)
+    writeFileSync(targetPath, file);
+  }
```
georgios-uber: Please write a message if not migrated and does not match!

PR #204, opened 2019-06-04
Remove styletron

jckr:
this removes styletron from the branch. this also adds gatsby-ssr.js to ocular's own website.


jckr:
this branch seems stable enough. using this approach on loaders.gl (with a gatsby-ssr.js file present) enabled us to fix examples issue. 
this doesn't fix styling issues long term though.

PR #207, opened 2019-06-06
Add new config DOC_FOLDERS to support multiple doc folders

javidhsueh:


modules/gatsby/src/gatsby-config/get-gatsby-config.js
```diff
@@ -323,6 +324,19 @@ module.exports = function getGatsbyConfig(config) {
 
   // conditional plug-ins - only added depending on options on config
 
+  if (paddedConfig.DOC_FOLDERS && paddedConfig.DOC_FOLDERS.length > 0) {
+    // Generates gatsby nodes for markdown files and JSON file in the in the docs folder
+    paddedConfig.DOC_FOLDERS.forEach(folderPath => {
+      gatsbyConfig.plugins.push({
+        resolve: 'gatsby-source-filesystem',
+        options: {
+          name: 'docs',
```
jckr: what does this do?

docs/gatsby/api-reference/ocular-config.md
```diff
@@ -18,6 +18,8 @@
 | `EXAMPLES`         | `Array` | See below |
 | `INDEX_PAGE_URL` | `String`  | Optional URL to a replacement component for the home page. |
 | `DOC_PAGE_URL` | `String`  | Optional URL to a replacement component for doc pages. |
+| `DOC_FOLDER` | `String`  | The path to the doc folder. |
```
jckr: right now DOC_FOLDER doesn't do anything

modules/gatsby/src/gatsby-config/get-gatsby-config.js
javidhsueh: It will source files in the folder into Gatsby system and create `File` nodes. Later on, `gatsby-transformer-remark` transforms markdown files into `MarkdownRemark` nodes.

Here I'm trying to source data from `DOC_FOLDER` or `DOC_FOLDERS`

docs/gatsby/api-reference/ocular-config.md
javidhsueh: But my changes will need it?

jckr:
from our conversation what this does is that this allows to build one single documentation tree from files existing in various folders. 
But we'll also want to build several documentation trees (ie not necessarily starting with /docs). 
What I'd like is an API that doesn't force us to change the configuration down the road.

ibgreen:
> But we'll also want to build several documentation trees (ie not necessarily starting with /docs).

What use case do you have in mind?

javidhsueh:
> from our conversation what this does is that this allows to build one single documentation tree from files existing in various folders.
> But we'll also want to build several documentation trees (ie not necessarily starting with /docs).
> What I'd like is an API that doesn't force us to change the configuration down the road.

Right now, it still supports 'DOC_FOLDER' and 'DOC_FOLDERS' at the same time. The 'DOC_FOLDER' will supersede over 'DOC_FOLDERS' if both of them exist.

docs/gatsby/api-reference/ocular-config.md
ibgreen: If DOC_FOLDER doesn't do anything let's delete it and use DOC_FOLDERS only? Having two of these does not seem to be in the user's interest.

modules/gatsby/src/gatsby-config/get-gatsby-config.js
ibgreen: Should these all have the same `name: 'docs'`? Does this affect how we query them?

modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
```diff
@@ -50,12 +50,24 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode(
   // Update path
   let relPath = node.fields.slug;
   if (node.fileAbsolutePath) {
-    const index = node.fileAbsolutePath.indexOf('docs');
-    if (index !== -1) {
-      relPath = node.fileAbsolutePath.slice(index);
+
+    const {ocularConfig} = global;
+    // TODO(@javidhsueh): we might deprecate DOC_FOLDER soon.
```
ibgreen: If it has no effect, let's just remove it. We obviously won't be breaking any web sites.

javidhsueh:
- Remove comment
 - Remove "DOC_FOLDER" from documentation
 - Add messages to warn that some config to be deprecated soon. (DOC_FOLDER)

jckr:
the use case I have in mind is vis academy which i think is not the only "classic" ocular website with several documentation trees. 
in vis academy, not only md files are in different folder, but they are built into different trees with different urls which don't necessarily start with /docs/.

PR #211, opened 2019-06-21
Do not add all files when linting

twojtasz:
This makes for a very unintuitive workflow where changes that have nothing to do with the actual code can be added as a result of linting.

If this is preferred as a simple choice of ocular, then I can try to add a flag/option to disable rather than remove it.


twojtasz:
@ib I'm guessing the `npx prettier --write` which will update docs might not be included if the flow immediately continued to a `commit`

PR #214, opened 2019-07-02
create base web shim

jckr:
so we can keep on using styled component with an API that looks like styletron. 

Ideally we'd use base-web and styletron. To that end we have two things to do:
- get styletron to work including server side and
- convert existing styling to base-web/styletron.

We could execute those tasks sequentially, ie: first get styletron to work as expected, THEN convert styling to styletron.
With this PR, we can do these in parallel. While we are figuring out how to get styletron to work, we can start converting existing styling to styletron, while these components are really styled by styled-components.

### DIFF
--- a/modules/gatsby/src/components/styled/index.js
+++ b/modules/gatsby/src/components/styled/index.js
@@ -1,11 +1,49 @@
+/* eslint-disable arrow-body-style */
+/* eslint-disable react/jsx-filename-extension */
 /* eslint-disable no-unused-vars */
-import styled from 'styled-components';
+import React from 'react';
+import styledComponents from 'styled-components';
+
 import theme from './theme';
 /* eslint-disable import/prefer-default-export */
 
+// baseweb shim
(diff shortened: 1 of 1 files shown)


modules/gatsby/src/components/styled/index.js
```diff
@@ -14,101 +52,102 @@ export const BodyContainerFull = styled.div(props => ({
   }
 }));
 
-export const BodyContainerToC = styled.div(props => ({
+
+export const BodyContainerToC = styled('div', ({$theme, ...props}) => ({
```
jckr: this is exactly how you'd write a base-web styled component. 

export const MyStyledComponent = styled('div', ({$theme, ...props}) => ({...}));

jckr:
This is a lowest common denominator approach. This doesn't support the full base web API, for instance we can't use props to the styled components. However, every styling function described like so will work with both base web / styletron and styled components with very high confidence. Once we do the switch there can be further light rework.

PR #216, opened 2019-07-10
Baseui fix

jckr:
This follows the pair programming session I had with @tajo. 
we are now importing styletron into gatsby through the gatsby-plugin-styletron which is the cleanest way. we can then add a BaseProvider wrapper. 

With this approach I've been able to get baseweb components working both on the development and the production version of ocular-gatsby. One condition though is that the local tree has the correct gatsby-browser.js and gatsby-ssr.js files:

gatsby-browser:
```
module.exports.wrapPageElement = require('@jckr/ocular-gatsby/gatsby-browser').wrapPageElement;
```

gatsby-ssr:
```
module.exports.wrapPageElement = require('@jckr/ocular-gatsby/gatsby-ssr').wrapPageElement;
```

### DIFF
--- a/modules/gatsby/package.json
+++ b/modules/gatsby/package.json
@@ -1,7 +1,8 @@
 {
-  "name": "ocular-gatsby",
+  "name": "@jckr/ocular-gatsby",
   "description": "GatsbyJS plugin for rendering markdown docs for frameworks.",
-  "version": "1.0.0-alpha.42",
+  "private": false,
+  "version": "1.0.0-alpha.43",
   "license": "MIT",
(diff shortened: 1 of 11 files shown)


modules/gatsby/package.json
```diff
@@ -84,6 +87,7 @@
     "slug": "^0.9.3",
     "stats.js": "^0.17.0",
     "styled-components": "^4.1.0",
+    "styletron-react": "^5.2.0",
```
tajo: this should not be needed since it's a depedency of `gatsby-plugin-styletron`

PR #224, opened 2019-08-08
Fixes to init process

jckr:
Hi, following the session with Isaac I wrote a number of small changes in the ocular-gatsby scripts and templates. 

1) First error - cannot find modules gh-pages
gh-pages now, it's possible to run ocular init without having gh-pages installed. 

2) I've reworded: 'Where is the ocular website relative to your main project?'

3) not sure what happened with .gitignore not copying, i couldn't repro

4) removing the copying of  /src/site-query.jsx.

5) our ocular-config.js didn't have a DIR_NAME in it
I touched up the template that generates it. I made sure to write projects with the right syntax (array vs object)

6) react-dom wasn't found
ok while react / react-dom should be peerDependencies of ocular-gatsby, they should be dependencies of the local tree, else we can go through the install process and not add them. so i added them in the deps of the template of the package.json

7) keeping license info
another problem I had written down and that I hadn't mentioned when reporting on our experience is that when Isaac created his original package.json file via npm init, he specified a license, but that license was then overwritten by ocular-gatsby. now, if a license is specified, it will be carried over during the installation.


modules/gatsby/bin/ocular.js
```diff
@@ -121,8 +137,17 @@ const commands = {
           const file = readFileSync(`${TEMPLATE_DIR}/${filename}`);
           console.log('Writing', `${DIR_PATH}/${filename}`, file.slice(20))
           writeFileSync(`${DIR_PATH}/${filename}`, file);
+        }        
+        const UPDATED_PACKAGE_JSON = require(`${DIR_PATH}/package.json`)
```
isaacbrodsky: Is it possible to just merge the new properties in without removing others that might already be present?

jckr: i may do a different PR for that.

PR #225, opened 2019-08-08
Generate table of contents

jckr:
This creates a script to generate a table of contents json file from the location where the documentations files are (as indicated in the ocular-config.js file)


jckr:
This is a pretty experimental process which I would like to test on our various ocular-gatsby sites. but the way it works is that it could be run any number of time and yes, it would overwrite the toc. 

so either - an author cares about having their toc in a certain way and manually curates it, or they don't and they can simply regenerate it whenever they modify content.

PR #226, opened 2019-08-09
Typography pass

jckr:
Screen Shot 2019-08-09 at 2 00 08 PM
Screen Shot 2019-08-09 at 1 59 58 PM

PR #227, opened 2019-08-09
Updated header

jckr:
Screen Shot 2019-08-09 at 3 55 16 PM


jckr:
Same as for #226 - this doesn't work well for mobile. i'm leaving this here as WIP and when we'll have a mobile implementation we can land it.

jckr:
I made some fixes to this pr to make responsive layout work. It's not perfect (i have not reimplemented TOC) but it works

Screen Shot 2019-08-12 at 11 06 32 PM
Screen Shot 2019-08-12 at 11 06 26 PM

PR #228, opened 2019-08-14
first stab at restyled table of contents

jckr:
Screen Shot 2019-08-14 at 10 53 10 AM
Screen Shot 2019-08-14 at 10 52 38 AM
Screen Shot 2019-08-14 at 10 52 26 AM

This is still WIP as there's a certain number of nits that I haven't ironed. for some reason the leaves links keep the color of the default link for instance. the toggle to expand/collapse toc for mobile view doesn't work as it used to (i just haven't explored it)

but this is how the code would look like. 
having toc entries with children that toggle when clicked, as opposed to link to the first entry, requires a complete overhaul of the toc code. but this time the same code works for regular toc and example toc.


modules/gatsby/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "ocular-gatsby",
+  "name": "@jckr/ocular-gatsby",
```
ibgreen: Change before landing.

modules/gatsby/package.json
```diff
@@ -1,8 +1,8 @@
...
   "description": "GatsbyJS plugin for rendering markdown docs for frameworks.",
   "private": false,
-  "version": "1.0.0-alpha.46",
+  "version": "1.0.0-alpha.49",
```
ibgreen: Bump version in separate commit.

modules/gatsby/src/components/styled/header.js
```diff
@@ -31,9 +31,9 @@ export const HeaderContainer = styled('div', ({$theme, ...props}) => ({
 }));
 
 export const HeaderMenuBlock = styled('div', ({$theme, ...props}) => ({
-  alignItems: 'center',
-  display: 'flex',
-  flexDirection: 'row'
+    alignItems: 'center',
+    display: 'flex',
+    flexDirection: 'row'
```
ibgreen: 4 space indentation? Check linter rules.

PR #231, opened 2019-08-16
Update markdown page styling

Pessimistress:
Sample: https://loaders.gl/docs/specifications/category-mesh

- Fix missing dependency
- Update markdown page styling
- Fix TOC link click area

### DIFF
--- a/modules/gatsby/src/components/layout/table-of-contents.jsx
+++ b/modules/gatsby/src/components/layout/table-of-contents.jsx
@@ -20,7 +20,7 @@
 
 import React, {PureComponent} from 'react';
 
-import {TocChevron, TocLink, TocEntry, TocSubpages} from '../styled';
+import {activeTocLinkStyle, TocChevron, TocHeader, TocLink, TocEntry, TocSubpages} from '../styled';
 
 export default class TableOfContents extends PureComponent {
   constructor(props) {
@@ -181,6 +181,7 @@ function updateHeights(tocEntries, expanded) {
 const SafeLink = ({
   depth,
   hasChildren,
+  isTocOpen,
   id,
   name,
   path,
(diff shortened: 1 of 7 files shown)


modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -196,11 +197,11 @@ const SafeLink = ({
 
   return (
     <TocEntry $depth={depth} title={name} onClick={() => toggleEntry(id)}>
-      {hasChildren && <TocChevron $depth={depth} />}
+      {hasChildren && <TocChevron $depth={depth} $isTocOpen={isTocOpen} />}
```
jckr: i see what you did there, I didn't implement rotating the toc chevron because design had it facing down whether the entry was open or not

modules/gatsby/src/components/styled/index.js
```diff
@@ -6,9 +6,16 @@ import ChevronDown from 'baseui/icon/chevron-down'
 import {styled} from 'baseui';
 /* eslint-disable import/prefer-default-export */
 
+import THEME from './theme';
```
jckr: we should refrain from using './theme'. it is a remnant from the time when i had trouble getting baseweb working server side. the problem with what is in ./theme is that it's static and won't be overridden by parameters that the user may pass. meanwhile, the actual baseweb theme can be accessed from any component via the provider pattern.

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -196,11 +197,11 @@ const SafeLink = ({
...
       {!path || typeof path !== 'string' ? (
-        <span>{name}</span>
+        <TocHeader $depth={depth}>{name}</TocHeader>
       ) : (
-        <TocLink $depth={depth} to={path} title={name}>
+        <TocLink $depth={depth} to={path} title={name} activeStyle={activeTocLinkStyle}>
```
jckr: i understand you try to style a link differently if it corresponds to the page displayed. 
 there's other ways we can do that w/o using activeStyle. renderRoute passes to SafeLink an "active" prop that can work, we just don't have it declared in SafeLink :|

PR #232, opened 2019-08-20
Post styling fixes

jckr:
This PR addresses styling issues where the style in production (gatsby build) is different from the style in development (gatsby develop).

we can't guarantee precedence for whatever styling is in a media query. 
so, in a code like this:

```
styled('div', ({$theme}) => ({
  height: '20px',
  [`@media screen and (max-width: ${$theme.breakpoints.medium}px)`]: {
    height: '40px',
  }
}));
```
we can't predict the actual height if the media query is triggered. it may be 20px or 40px and it may be different in prod or dev.

The solution is to put all statements that could be affected by a media query in an opposite, mutually exclusive media query, ie:

```
styled('div', ({$theme}) => ({
  [`@media screen and (min-width: ${$theme.breakpoints.medium}px)`]: {
    height: '20px',
  }
  [`@media screen and (max-width: ${$theme.breakpoints.medium}px)`]: {
    height: '40px',
  }
}));
```
now we are sure that the this element will either be 20px high in large displays, and 40px high in small displays. 

In addition, this PR introduces a styled table with margin before and after and styling consistent with baseweb.

Screen Shot 2019-08-19 at 7 18 59 PM

### DIFF
--- a/modules/gatsby/src/components/styled/header.js
+++ b/modules/gatsby/src/components/styled/header.js
@@ -11,11 +11,13 @@ export const Header = styled('header', ({$theme}) => ({
   height: $theme.sizing.scale1600,
   justifyContent: 'space-between',
   padding: `0 36px`,
-  position: 'fixed',
   top: 0,
   left: 0,
   width: '100%',
   userSelect: 'none',
+  [`@media screen and (min-width: ${$theme.breakpoints.medium}px)`]: {
+    position: 'fixed',
+  },
   [`@media screen and (max-width: ${$theme.breakpoints.medium}px)`]: {
     position: 'static'
   }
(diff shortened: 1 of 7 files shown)


jckr:
ok @Pessimistress we have lots of overlaps between our 2 PRs

modules/gatsby/src/components/styled/header.js
```diff
@@ -11,11 +11,13 @@ export const Header = styled('header', ({$theme}) => ({
   height: $theme.sizing.scale1600,
   justifyContent: 'space-between',
   padding: `0 36px`,
-  position: 'fixed',
   top: 0,
   left: 0,
   width: '100%',
   userSelect: 'none',
+  [`@media screen and (min-width: ${$theme.breakpoints.medium}px)`]: {
```
ibgreen: Worth adding comments with the explanation you gave in the PR?

jckr: I will abandon this PR though. As discussed we should merge #231, then have a PR that only focuses on these media query issues.

jckr:
this is replaced by #231 + #233

Issue #229, opened 2019-08-15
Internal link gets replaced with `undefined` or requires odd pathing


isaacbrodsky:
Please see uber/h3#274

There were two issues I had with `a href` links. The first is that when an internal link wasn't resolved, it was replaced with `undefined`. This silently broke the site, but really should have failed the build. I see the code for Markdown pages has changed in master so I'm not sure if this issue still exists.

The second issue was that for the page `/docs/core-library/filters.md` to reference `/docs/core-library/h3Indexing.md`, I need to write the link as `../h3indexing.md` (note the `..` and that the capitalization changed). I think this may be an issue with documents nested more than one level deep since I didn't see this on the React Vis site. I would have preferred either `h3Indexing.md` or `/docs/core-library/h3Indexing.md`.

isaacbrodsky:
Is there a recommended way to add internal links in ocular-gatsby? @jckr

Issue #230, opened 2019-08-15
Image prefixing not working


isaacbrodsky:
Please see uber/h3#274

I couldn't find a way to use relative paths (such as `ijkp.png`, relative to the images directory, or `/images/ijkp.png`, relative to the site root) to refer to images. Instead I had to hardcode the path prefix, which results in breaking the development version of the site.

jckr:
please put your images in /src. Gatsby will transform them and apply various optimizations when creating the production website. That's why logically, it makes sense to have images along other parts of the source code. 

With your images in /src (for instance in /src/images) you can then reference them like so in your markdown: ![optional alt text]( path to your image from /src) 
for instance ![](/images/ijkp.png)

ibgreen:
Should we add a chapter about images in our docs and mention this there?

isaacbrodsky:
@jckr I assumed `static/images` (rather than `src/images`) was the correct path because `npm run build` fails if it doesn't exist.

I tried both:

```
![IJK+](/images/ijkp.png)
![IJK+](images/ijkp.png)
```

and the resulting `img` tag did not have the prefix applied. The file was at `website/src/images/ijkp.png` in my repo.

PR #234, opened 2019-08-21
fix relative links in same physical folder

jckr:
This is in response to #229.


jckr:
Hi @isaacbrodsky, this addresses #229. I also left a comment on #230 so lmk if that approach works for you 
Screen Shot 2019-08-21 at 3 43 02 PM

jckr:
I've started a branch with updated documentation. we definitely need more tests too

PR #237, opened 2019-08-30
ocular-gatsby@1.0.0

ibgreen:


modules/gatsby/src/components/styled/index.js
```diff
@@ -81,7 +81,7 @@ export const BodyGrid = styled('div', ({$theme, ...props}) => ({
   display: 'grid',
   gridTemplateRows: '64px 1fr',
   gridTemplateColumns: '300px 1fr',
-  maxWidth: `${$theme.breakpoints.large}px`,
+  // maxWidth: `${$theme.breakpoints.large}px`,
```
jckr: we can just delete it. we don't want it back

ibgreen: deleted

ibgreen:
1.0.0 published \o/

PR #249, opened 2019-09-18
fix layout issues (see luma.gl)

jckr:
The point of this PR is to ensure we have scroll conservation when we change routes. 
Currently, when a user clicks on a link in the TOC, in the new page both the TOC and main page scroll to top. 
To prevent that, instead of a layout using display: grid, we:
- have the TOC displayed using fixed and positioned right under the header. we also give a zindex so it can be displayed over the rest of the page;
- add padding to the main part of the screen. the left padding corresponds to the toc width. 
As a result, both parts scroll independently and scrolling position is conserved when changing pages. 

The behavior for smaller screens is conserved, ie: when = table of contents is clicked, the toc will replace the documentation. 

However, the small screens had an additional problem that this PR fixes, which is the behavior of the project links menu. Previously, this menu will appear under the "table of contents" toggle, and would stay on when scrolling down the page. (fixed position). 

Instead, I implement the design, which is: when we click on the hamburger menu, the list of projects takes over the main page. The implementation of that had to be slightly different for pages with TOC and pages without. 

scroll conservation - full - small
scroll conservation - full - wide

### DIFF
--- a/modules/gatsby/src/components/layout/header.jsx
+++ b/modules/gatsby/src/components/layout/header.jsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-array-index-key */
 // Copyright (c) 2018 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -35,6 +36,9 @@ import {
   HeaderMenu,
   HeaderMenuLink
 } from '../styled/';
+import {navigate} from 'gatsby';
+import {Search} from 'baseui/icon';
+import {Input} from 'baseui/input';
 
 import GithubStars from '../github/github-stars.jsx';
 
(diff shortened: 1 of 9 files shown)


modules/gatsby/src/components/layout/header.jsx
```diff
@@ -35,6 +36,9 @@ import {
   HeaderMenu,
   HeaderMenuLink
 } from '../styled/';
+import {navigate} from 'gatsby';
+import {Search} from 'baseui/icon';
+import {Input} from 'baseui/input';
 
 import GithubStars from '../github/github-stars.jsx';
```
ibgreen: Unrelated... any chance we can get github stars back?

modules/gatsby/src/components/layout/header.jsx
```diff
@@ -98,6 +102,94 @@ function generateHeaderLinks(props) {
   return links.filter(Boolean);
 }
 
+const ControlledHeader = ({
```
ibgreen: Nit: One file per component?

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -47,7 +47,15 @@ export default class TableOfContents extends PureComponent {
     this.toggleEntry = this.toggleEntry.bind(this);
   }
 
+
+  componentDidMount() {
+    console.log('CDM');
```
ibgreen: Drop? 

Or import probe.gl so that we can enable ocular debug logging conditionally.

Right now there are lots of warnings in ocular console that are quite distracting...

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -47,7 +47,15 @@ export default class TableOfContents extends PureComponent {
...
+  }
+  componentWillUnmount() {
+    console.log('CWU');
```
ibgreen: Drop

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -47,7 +47,15 @@ export default class TableOfContents extends PureComponent {
...
+  }
   componentDidUpdate(prevProps) {
+    console.log('CDU');
```
ibgreen: Drop

modules/gatsby/src/components/layout/table-of-contents.jsx
```diff
@@ -79,6 +87,7 @@ export default class TableOfContents extends PureComponent {
   }
 
   render() {
+    console.log('R');
```
ibgreen: Drop

modules/gatsby/src/components/layout/header.jsx
jckr: oh, i have to remove those extra lines too. I was experimenting with having a search bar in the top header. 

we'll need to integrate that in the design. Design of the home page has a lot of room for improvement. The header is ok but home page is a mess.

modules/gatsby/src/components/layout/table-of-contents.jsx
jckr: thanks

PR #251, opened 2019-09-20
tweak build to accept options to build selected targets.

kenns29:
tweak ocular-build to accept options to build selected targets. Addressing this issue https://github.com/uber/manifold/issues/31.

**ocular-build**

Usage:

```
ocular-build [--dist es5|es6|esm,...] [module,...]
```

Example:

```
ocular-build --dist es6,esm
ocular-build --dist esm
ocular-build --dist es6,esm module1
ocular-build module1
```

New Behaviors:
1. If there's only one option after `--dist`, then the bundle will be built into `dist/` (without `esm/` etc .)
2. If there are unrecognized parameter after the `--dist` option, ocular will give an error message and exit.
3. The pattern with `\-\-[a-zA-Z]+` after `ocular-build` is now reserved for options, in the other words, there can't be a module named as these, e.g. `--m` as module...

**ocular-clean**

Usage

```
ocular-clean [all]
```

New Behavior:
1. If the `all' option is used, clean the entire dist/ without making empty es5/ es6/ esm/

### DIFF
--- a/modules/dev-tools/scripts/build.sh
+++ b/modules/dev-tools/scripts/build.sh
@@ -7,7 +7,7 @@ CONFIG=`node $DEV_TOOLS_DIR/node/get-config.js ".babel.configPath"`
 
 check_target() {
   if [[ ! $1 =~ ^es5|es6|esm ]]; then
-    echo "Build target $1 not recognized."
+    echo -e "\033[91mUnknown build target $1. ocular-build [-t es5|es6|esm,...] [module1,...]\033[0m"
     exit 1
   fi
 }
@@ -44,7 +44,7 @@ build_monorepo() {
             TARGET=$2
             shift ;;
         *)
-            echo "Option $1 not recognized."
+            echo -e "\033[91mUnknown option $1. ocular-build [-t es5|es6|esm,...] [module1,...]\033[0m"
             exit 1 ;;
       esac
     else
--- a/modules/dev-tools/scripts/build.sh
+++ b/modules/dev-tools/scripts/build.sh
@@ -58,7 +58,7 @@ build_monorepo() {
 
   cd modules
 
-  if [ -z "$MODULES"]; then
+  if [ -z "$MODULES" ]; then
     # Build all modules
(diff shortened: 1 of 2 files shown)


kenns29:
The ocular-clean command only cleans the es5/ es6 esm/ folders but not the entire dist/, if I add the option to build stuff into dist/ itself, then I the clean will have to clean everything in dist/. Is it OK if I tweak ocular to clean dist/*, otherwise, the implementation gets messier ...

kenns29:
> The ocular-clean command only cleans the es5/ es6 esm/ folders but not the entire dist/. Is it OK to let ocular-clean clean everything in dist/*, otherwise, the implementation gets messier ...

Added a `-a' option in ocular-clean to address this problem.

kenns29:
> General comments:
> 
> * There are multiple complex improvements here (dist targets, module selection, clean), might be better with smaller focused PRs.
> * Ocular sub-commands/options are generally full words, rather than hyphenated options (compare `ocular-lint fix`, not `ocular-lint -f`)
> * Changing this system to support module names at the end doesn't feel great. If we have to change, at least lets use long options (`--fix`, `--dist` instead of `-f`, `-t`). But I think instead we can just define the commands to accept a module list.
> * We should add `ocular-*** help` command to all ocular scripts (already supported by some) and list the new options/commands there.
> 
> > The ocular-clean command only cleans the es5/ es6 esm/ folders but not the entire dist/. Is it OK to let ocular-clean clean everything in dist/*, otherwise, the implementation gets messier ...
> 
> I think so, however some operations could crash if the subdirectories don't exist, so we just need to make sure we always create them first. We should find this pretty quickly in testing with some of the existing repos, you just need to be ready to do quick fixes.
> 
> Another thought is, what should ocular-clean actually clean? Sometimes I like to clean out all node-modules as well and get back to pristine state. So `clean all` or `clean install` could be some nice options...

Changed the options based on the comments. Btw, this PR only has only added the dist targets and clean all option, the module selection option has been in ocular-build before this PR.

PR #258, opened 2019-09-26
Improve pre-commit lint script

Pessimistress:
- Block commit if prettier check does not pass, instead of silently modifying files
- Only run prettier & eslint on files that are in `config.lint.paths`
- Run all files together instead of one-by-one, so that all errors are printed

### DIFF
--- a/modules/dev-tools/scripts/lint.sh
+++ b/modules/dev-tools/scripts/lint.sh
@@ -39,19 +39,25 @@ case $MODE in
   "pre-commit")
     print_yellow "Running prettier & eslint on changed files..."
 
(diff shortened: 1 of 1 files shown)


modules/dev-tools/scripts/lint.sh
```diff
@@ -39,19 +39,25 @@ case $MODE in
   "pre-commit")
     print_yellow "Running prettier & eslint on changed files..."
 
+    NAME_PATTERN=`echo "^$DIRECTORIES/.*\.js$" | sed -e 's/,/|/g' | sed -e 's/{/(/g' | sed -e 's/}/)/g'`
```
kenns29: Curious why you have to do this:
`sed -e 's/,/|/g' | sed -e 's/{/(/g' | sed -e 's/}/)/g'`

Pessimistress: I'm basically translating the glob pattern to a regex pattern.

PR #270, opened 2019-12-07
gatsby-theme-ocular: repackage ocular-gatsby as a gatsby theme

ibgreen:
With a surprisingly small amount of changes this repackages `ocular-gatsby` into `gatsby-theme-ocular`.

There are many big upsides to this change:
- Ease of use: A theme can be installed by a single line in any gatsby site's `gatsby-config.js`.
- Maintainability: Adopting the new theme system means we are now using officially an supported integration mechanism.
- Workspaces: Adopts a gatsby theme recommended workspace setup, which allows for the ocular website to be used as a test bed without any tricky linking requirements.

A quick pass on the docs was made to update text and cover upgrade steps.

I'd expect websites to quickly adapt this approach, but to support any hotfixes on the `ocular-gatsby` module I have pushed a `release-v1-ocular-gatsby` branch that can be used for maintenance.


README.md
```diff
@@ -1,5 +1,8 @@
 # ocular
 
-Ocular is a tool primarily designed for building documentation websites for github-based javascript frameworks, built using the gatsbyjs documentation generation system.
+Ocular is a monorepo for development tools primarily designed for building github-based javascript frameworks.
+
+It currently contains:
+- `gatsby-theme-ocular` - A documentation generator packaged as a pluggable theme for gatsbyjs documentation generation system.
+- `ocular-dev-tools` - A set tools that packages up extensive installation and configurationof babel, webpack, lint, prettier and other state-of-the art JS build tools.
```
tgorkin: Minor typo, missing a space in `configurationof`.

README.md
```diff
@@ -1,5 +1,8 @@
 # ocular
 
-Ocular is a tool primarily designed for building documentation websites for github-based javascript frameworks, built using the gatsbyjs documentation generation system.
+Ocular is a monorepo for development tools primarily designed for building github-based javascript frameworks.
```
tgorkin: With this major change to move towards using gatsby themes,  it worth considering moving gatsby-theme-ocular out of the ocular monorepo?  I don't see lots of overlapping code or scripts between dev tools and ocular-gatsby, and the move to themes seems to further diverge the components.  Maybe too much work at this point, but worth a discussion.

docs/get-started.md
```diff
@@ -17,11 +19,11 @@ npm init -y
...
-yarn add -D ocular-gatsby
+yarn add -D gatsby-theme-ocular
 ```
 or
 ```
-npm install ocular-gatsby --save-dev
+npm install gatsby-theme-ocular --save-dev
 ```
 
 ## Creating and running your Ocular website
```
tgorkin: Are these instructions about running `ocular init` still correct?  Looks like ocular-config.js is still necessary but many options have been moved into gatsby-config.js.

I think these getting started instructions could probably use more elaboration. Specifically would be good to briefly cover the necessary config files that are created, along with links to relevant https://www.gatsbyjs.org/ doc pages.

modules/gatsby-theme-ocular/bin/ocular.js
```diff
@@ -146,7 +146,7 @@ const commands = {
         }
 
         writeFileSync(`${DIR_PATH}/package.json`, `${JSON.stringify(UPDATED_PACKAGE_JSON, null, 2)}\n`)
-        
+
         const ocularConfig = OCULAR_CONFIG_TEMPLATE(result);
         writeFileSync(`${DIR_PATH}/ocular-config.js`, ocularConfig);
```
tgorkin: Can all of these commented out writeFileSync calls be deleted?

modules/gatsby-theme-ocular/bin/ocular.js
```diff
@@ -136,7 +136,7 @@ const commands = {
           const file = readFileSync(`${TEMPLATE_DIR}/${filename}`);
```
tgorkin: Can we delete the commented out code above?  Would be good to clean this file up to make it less confusing, unless these commented out snippets are critical for debugging.

modules/gatsby-theme-ocular/gatsby-config.js
```diff
@@ -0,0 +1,3 @@
+const getGetsbyConfig = require('./src/gatsby-config/get-gatsby-config');
```
tgorkin: Typo Getsby => Gatsby

README.md
ibgreen: >  Is it worth considering moving gatsby-theme-ocular out of the ocular monorepo? 

Yes, though if anything should be moved out to another repo, it is probably the `dev-tools` module, since the `ocular` name (and this repo) was originally created for the first (non-gatsby) version of our documentation generator.

An argument for keeping them here is that each repo comes with a certain amount of overhead. While code sharing is limited, these modules both follows a theme of being developer automation type tools, and one could imagine it being a home for additional automation tools in the future.

Also, a big push in the recent v8 time frame has been to consolidate into fewer repos, examples:
- `viewport-mercator-project` went into `@math.gl/web-mercator`
- `seer` went into `@probe.gl/seer`
- `monochrome` into `streetscape.gl` 
- etc.

ibgreen:
> Also on the point about Workspaces, curious if that is usable outside of this repository? This seems like a really nice improvement to development workflow.

Yes, per my tests, if:
- both an example folder and and a "library module" (dependency of the example) are part of the workspaces
- and the example referencing the module with a `*` version
- and the module's package.json specifies a not-yet-published version number
Then the example's references to that module will automatically resolve to the local version.

Currently we typically don't make examples part of the workspaces setup, so that would be the major change.

> Overall conditional approval from me, representing the vis framework websites as customers. However we should get consensus from the other gatsby maintainers.

Sounds good.

ibgreen:
> Anything the old approach might be able to do that a theme wouldn't be able to?

Probably, as the old approach took control of everything and had the last word on the generated config, there could now be cases where we'd not be able to override other plugins and themes. 

But in the previous version it was very hard for the user to apply other plugins and themes in the first place, so I would argue the situation is much better, and we are on a level playing ground with all other modules.

tgorkin:
> Probably, as the old approach took control of everything and had the last word on the generated config, there could now be cases where we'd not be able to override other plugins and themes.

To clarify, can we still add our own project-specific plugins? Just not override plugin settings defined down in the theme?  For example, I added the 'gatsby-plugin-sass' for the deck.gl gatsby website.  Curious if this is still possible.

ibgreen:
> To clarify, can we still add our own project-specific plugins? 

In the latest PR, there is no longer a need for ocular-config.js. You just have a top-level gatsby-config.js you can add plugins next to `ocular-theme-gatsby` in the plugins list

https://github.com/uber-web/ocular/blob/ib/improve-markdown-title-extraction/website/gatsby-config.js

If you need to inject plugins in the middle of the ocular injected plugins we need to provide a bit more control of course, but most plugins should be fine to use before or after.

modules/gatsby-theme-ocular/bin/ocular.js
ibgreen: deleted

docs/get-started.md
ibgreen: > Are these instructions about running ocular init still correct? 

I patched them up somewhat, but I think unless someone steps in and adds some test cases to `ocular init` we should probably remove it - most websites are created as a copy paste exercise, not with `ocular init`, and now ocular is just one of many plugins, better run something like `gatsby init`...

ibgreen:
merging into feature branch

Issue #268, opened 2019-12-06
Doc entries disappearing in the luma.gl site 


tsherif:
The site is staged here: https://tsherif.github.io/luma.gl/docs

Some entries in the table of contents aren't showing up in the generated sidebar, e.g. the first entry in the API reference section is a submenu "Overview" with README titled "Overview":

<img width="636" alt="Screen Shot 2019-12-06 at 8 46 59 AM" src="https://user-images.githubusercontent.com/476017/70327629-387f6e80-1805-11ea-85fc-d8f766400eee.png">

But it doesn't show up in the sidebar:
<img width="350" alt="Screen Shot 2019-12-06 at 8 46 46 AM" src="https://user-images.githubusercontent.com/476017/70327758-7e3c3700-1805-11ea-8a62-8a71bc217150.png">

It is however accessible via direct navigation: https://tsherif.github.io/luma.gl/docs/api-reference

A few other entries that aren't listed:
- https://tsherif.github.io/luma.gl/docs/get-started
- https://tsherif.github.io/luma.gl/docs/contributor-guide

It seems to be only files named `README.md`, but changing the names didn't help.

ibgreen:
There are some particularities in ocular's processing of markdown.
-  `README.md` files are treated as the default entry for that directory. If you delete the `/README` part it should work. 
- There must be a `# <title>` markdown stanza on the first line of the file

I did this in the PR just opened and it seems to have fixed it.

While easy to fix, there should be better defaults and reasonable warnings for these cases.

PR #273, opened 2019-12-09
better treatment of readme.md files

jckr:
This is related to https://github.com/uber-web/ocular/issues/268. 
If the TOC was parsed before markdown files, we try to find the corresponding entry to a markdown file in the TOC as we go. We were doing that by matching the relPath variable to the entry property of a toc node. These values typically match but for a readme.md node the syntax of the relPath is going to be slightly different. This PR solves this problem.

### DIFF
--- a/modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
+++ b/modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
@@ -29,7 +29,7 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode(
   tocNode
 ) {
   const {createNodeField} = actions;
-
+debugger;
   const fileNode = getNode(node.parent);
   const parsedFilePath = path.parse(fileNode.relativePath);
   const hasTitle =
@@ -49,6 +49,7 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode(
 
   // Update path
   let relPath = node.fields.slug;
+  let entry = relPath;
   if (node.fileAbsolutePath) {
 
     const {ocularConfig} = global;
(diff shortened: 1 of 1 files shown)


modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
```diff
@@ -29,7 +29,7 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode(
   tocNode
 ) {
   const {createNodeField} = actions;
-
+debugger;
```
ibgreen: remove...

modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
```diff
@@ -49,6 +49,7 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode(
 
   // Update path
   let relPath = node.fields.slug;
+  let entry = relPath;
```
ibgreen: Does entry ever change? Should it be `const`? Does our test setup not run linter?

modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
```diff
@@ -69,7 +70,8 @@ module.exports.processNewMarkdownNode = function processNewMarkdownNode(
 
     const basename = path.basename(relPath, '.md');
     const dirname = path.dirname(relPath);
-    relPath = basename === 'README' ? dirname : `${dirname}/${basename}`;
+    entry = `${dirname}/${basename}`;
+    relPath = basename === 'README' ? dirname : entry;
 
     createNodeField({node, name: 'path', value: relPath});
     createNodeField({node, name: 'slug', value: relPath});
```
ibgreen: Is it better to determine the path when we are done matching?

modules/gatsby/src/gatsby-node/process-nodes/process-nodes-markdown.js
jckr: oops thanks!

jckr:
Indeed. There is a lot of rewriting which can probably be avoided in the first place.

PR #274, opened 2019-12-09
gatsby-theme-ocular: Improve markdown header extraction

ibgreen:
This is for kepler.gl. kepler.gkl uses `documentation.js` to generate markdown, which generates HTML comments before the markdown header - something the previous simple title parser in `gatsby-theme-ocular` couldn't handle.

This change looks for the first markdown title even if it is not on the first line or a single `#`.

Should also help with deck.gl which has some code injection tokens before titles in some markdown files.


modules/gatsby-theme-ocular/gatsby-node.js
```diff
@@ -1,25 +1,11 @@
...
 const sourceNodes = require('./src/gatsby-node/source-nodes');
+const setFieldsOnGraphQLNodeType = require('./src/gatsby-node/set-fields-on-graphgl-node-type');
 const onCreateNode = require('./src/gatsby-node/on-create-node/on-create-node');
 const onCreateWebpackConfig = require('./src/gatsby-node/on-create-webpack-config/on-create-webpack-config');
 const createPages = require('./src/gatsby-node/create-pages/create-pages');
 
-// TODO - move to the src directory
-const {addSiblingNodes} = require('./src/gatsby-node/on-create-node/process-nodes-markdown');
-
-function onPreBootstrap({ reporter }, ocularOptions) {
```
ibgreen: This is just cleanup, moving into the source tree in the right place.

PR #280, opened 2019-12-16
dont touch links if a relative link to another doc is not found

jckr:


modules/gatsby-theme-ocular/src/templates/doc-page-markdown.jsx
```diff
@@ -34,7 +34,7 @@ const CustomLinkWrapper = relativeLinks => {
       return <A href={href} {...props} />;
     }
     const hrefWithoutLeadingSlash = href.startsWith('/') ? href.slice(1) : href;
-    return <A {...props} href={relativeLinks[hrefWithoutLeadingSlash]} />;
+    return <A {...props} href={relativeLinks[hrefWithoutLeadingSlash]} || href/>;
```
ibgreen: The fix is fine, however from a maintainability perspective, stashing behavior that requires a PR to fix inside an inline expression without comments is pretty subtle and not very self-documenting.

How about a helper function 

```js
function resolveLink(href) {
   const hrefWithoutLeadingSlash = href.startsWith('/') ? href.slice(1) : href;
   // dont touch links if a relative link to another doc is not found
   return relativeLinks[hrefWithoutLeadingSlash]} || href;
}
```

even better:
-  start extracting that function to a helper file `parse-links.js` or similar.

jckr:
@ibgreen I have put all the links util functions (on the doc creation side and the ones we refer to in the first version of that PR) into the same util file as suggested.

ibgreen:
>  have put all the links util functions (on the doc creation side and the ones we refer to in the first version of that PR) into the same util file as suggested.

\o/ This is excellent! It will make it so much easier for contributors since it separates the link processing from the gatsby page generation and the react rendering!

jckr:
@ibgreen what's the process for versioning? I see there are several alpha versions of 1.1.0 on npm

ibgreen:
@jckr Yes, this is a good time to go to official 1.1.0. 

Keep the change log / what's new / docs updated?

PR #284, opened 2019-12-23
Fix bump script to detect peerDependencies

xintongxia:


modules/dev-tools/scripts/bump.js
```diff
@@ -39,6 +40,10 @@ function bumpPackages(packages) {
         devDependencies[p.name] = `^${p.version}`;
         changed = true;
       }
+      if (peerDependencies[p.name]) {
+        peerDependencies[p.name] = `^${p.version}`;
```
ibgreen: Nit: `p` => `package` ;)

PR #286, opened 2020-01-09
Fix eslint error when using 'ocular-lint fix'

Pessimistress:


xintongxia:
Is it intended to only fix `js` files?

Pessimistress:
This is eslint, not prettier. Only js is supported.

PR #287, opened 2020-01-11
dev-tools: Bump es6 build targets to Node 10

ibgreen:
Node 8 is no longer supported (LTS expired). This reduces the amount of transpilation in es6 mode enabling better debugging and performance.

### DIFF
--- a/modules/dev-tools/config/babel.config.js
+++ b/modules/dev-tools/config/babel.config.js
@@ -8,7 +8,7 @@ const ES6_TARGETS = {
   firefox: '60', // Released: 2018-May-9, https://en.wikipedia.org/wiki/Firefox_version_history
   safari: '12', // Released: 2018-09-07 (OSX Mojave) - https://en.wikipedia.org/wiki/Safari_version_history
   ios: '12', // Track Safari
-  node: '8' // Node 8 is LTS until December 31, 2019.
+  node: '10' // Node 8 LTS expired December 31, 2019.
 };
 
 const COMMON_CONFIG = {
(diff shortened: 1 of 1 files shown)


modules/dev-tools/config/babel.config.js
```diff
@@ -8,7 +8,7 @@ const ES6_TARGETS = {
   firefox: '60', // Released: 2018-May-9, https://en.wikipedia.org/wiki/Firefox_version_history
   safari: '12', // Released: 2018-09-07 (OSX Mojave) - https://en.wikipedia.org/wiki/Safari_version_history
   ios: '12', // Track Safari
-  node: '8' // Node 8 is LTS until December 31, 2019.
+  node: '10' // Node 8 LTS expired December 31, 2019.
```
Pessimistress: I think a more robust target would be `maintained node versions`, which is currently >=10. The rest of the browsers could be targeted with `last 2 versions`, which covers >87% of all users.

Reference:

[browserlist documentation](https://github.com/browserslist/browserslist#full-list)
[browserlist query online](https://browserl.ist/)

ibgreen: That sounds reasonable for node.js, and Chrome/Firefox. 

However, the targets for the smaller browsers were carefully selected to avoid pulling in unnecessary transforms. Relaxing these targets dramatically increases the amount of transpilation. 

Right now I rely heavily on the `es6` dist is to enable a sane debug experience with new constructs like async/await, and that is dependent on those browser versions. 

I will update and add comments around this.

Also, it could be time to think through our selection of distributions:
- Do we still need the es5 dist? Is there any bundler that actually uses it? Or make the esm dist our `main` dist?
- Could we stop transpiling certain features in our `esnext` dist? (e.g. math.gl gets a major performance boost when we disable the class transform).
- The current es6 dist is now really an `es2020` dist, maybe time to rename it to something less version specific?

ibgreen: Looked at existing comment header, it is pretty clear about why targets are specified the way they are. Keeping as is for now.

PR #300, opened 2020-02-25
Fix doc page redefinition

ibgreen:
The doc page template cannot be copied and modified since it imports some unexported componentry. 
- Supporting this enables math functions to be fixed which would remove the final known blocker for H3 adoption of `gatsby-theme-ocular` 
- (confirmed by editing the doc template in node_modules).

### DIFF
--- a/docs/gatsby-theme-ocular/api-reference/options.md
+++ b/docs/gatsby-theme-ocular/api-reference/options.md
@@ -29,7 +29,7 @@ The following options are available:
 | `EXAMPLES`         | `Array` | See below |
 | `STYLESHEETS`      | `Array` | Additional CSS stylesheets |
 | `INDEX_PAGE_URL` | `String`  | Optional URL to a replacement component for the home page. |
-| `DOC_PAGE_URL` | `String`  | Optional URL to a replacement component for doc pages. |
+| `DOC_MARKDOWN_PAGE_URL` | `String`  | Optional URL to a replacement component for doc pages. |
 | `DOC_FOLDERS` | `Array`  | A set of paths to the doc folders to source all markdown files. |
 | `EXAMPLE_GALLERY_PAGE_URL` | `String`  | Optional URL to a replacement component for the example gallery page (`/examples`). |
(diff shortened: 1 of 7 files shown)


docs/gatsby-theme-ocular/api-reference/options.md
```diff
@@ -29,7 +29,7 @@ The following options are available:
 | `EXAMPLES`         | `Array` | See below |
 | `STYLESHEETS`      | `Array` | Additional CSS stylesheets |
 | `INDEX_PAGE_URL` | `String`  | Optional URL to a replacement component for the home page. |
-| `DOC_PAGE_URL` | `String`  | Optional URL to a replacement component for doc pages. |
+| `DOC_MARKDOWN_PAGE_URL` | `String`  | Optional URL to a replacement component for doc pages. |
```
ibgreen: This just reflects what was working, `DOC_PAGE_URL` support was broken as far as I could tell.

website/package.json
```diff
@@ -13,9 +13,10 @@
     "serve": "gatsby serve --prefix-paths",
     "deploy": "NODE_DEBUG=gh-pages gh-pages -d public"
   },
+  "dep_comment": "Ensure gatsby-theme-ocular references unpublished version to trigger local workspaces link",
   "dependencies": {
     "gatsby": "^2.18.0",
-    "gatsby-theme-ocular": "*",
+    "gatsby-theme-ocular": "1.2.0-beta.8",
```
ibgreen: This triggers local workspaces install, the downside is that it has to be bumped both here and in the module to a matching unpublished version.

isaacbrodsky:
This still requires some fix ups in the copied file I believe, maybe we could add some comments/documentation on how exactly to do that?

ibgreen:
> This still requires some fix ups in the copied file I believe, maybe we could add some comments/documentation on how exactly to do that?

Good point. A bit too much work for this PR but opened an issue https://github.com/uber-web/ocular/issues/301.

PR #312, opened 2020-06-22
Gatsby: support MDX

Pessimistress:
- Support [MDX](https://www.gatsbyjs.org/docs/mdx/getting-started/) (JSX in markdown)
- Add `PAGES` to config to build arbitrary pages
- Allow `EXAMPLES` to link to external pages
- Fix broken link in search results


modules/gatsby-theme-ocular/src/gatsby-config/get-gatsby-config.js
```diff
@@ -74,46 +110,16 @@ module.exports = function getGatsbyConfig(config) {
       'gatsby-plugin-sharp',
       'gatsby-transformer-sharp',
 
-      // Transforms markdown (.md) nodes, converting the raw markdown to HTML
+      // Register plugins used by MDX
+      ...remarkPlugins,
```
ibgreen, line 114: - These were nested plugins for transformer-remark, are they now OK to have on top level? Do they actually do anything there?
- Or is MDX plugin completely taking over from the remark transformer?
- Or is the comment correct, they need to be listed twice?? 

Worth being a bit more verbose in the comment?

modules/gatsby-theme-ocular/src/gatsby-node/create-pages/page-templates.js
```diff
@@ -0,0 +1,35 @@
...
+const EXAMPLE_GALLERY_PAGE_URL = path.resolve(
+  templatePath,
+  'examples.jsx'
+);
+
+const SEARCH_PAGE_URL = path.resolve(templatePath, 'search.jsx');
+
+// All templates are overridable by top-level fields in ocular config
+module.exports = {
+  INDEX_PAGE_URL,
```
ibgreen, line 27: Maybe just inline the definitions here, no need for intermediates?

modules/gatsby-theme-ocular/src/gatsby-config/get-gatsby-config.js
Pessimistress: `gatsby-plugin-mdx` replaces `gatsby-transformer-remark`:

https://www.gatsbyjs.org/docs/mdx/migrate-remark-to-mdx/#replacing-gatsby-transformer-remark-with-gatsby-plugin-mdx

The plugin requires all sub-plugins be registered twice:

https://www.gatsbyjs.org/docs/mdx/plugins/

PR #331, opened 2020-10-07
Gatsby - Add codesandbox plugin

mletic:
Through this PR I've added `gatsby-remark-embedded-codesandbox` plugin that allows us to use [CodeSandbox](https://codesandbox.io/) examples that are stored locally.


ibgreen (approved):
Don't forget to update the ocular docs.

Also there is place where we can declare the GraphQL schema so that gatsby doesn't crash due to failed graphql schema auto-inference if this field is not supplied.

ibgreen:
@mletic Please add @Pessimistress to ocular code reviews.

mletic:
I plan to enhance the ocular docs in a different PR, as there are more props in the config of the theme that don't have explanations on what they are and how they are used.

ibgreen (approved):
I think we need some kind of paragraph in a guide page on how this works, not just a single line in the API reference.

Having all examples in one folder seems a little restrictive but I suppose this is a limitation of the plugin.

Have you confirmed that this is the same approach as in nebula.gl?

mletic:
Agreed on the documentation. I will expand with real examples on how this works.
Regarding one folder... You can always set the root of the project as the root for codesandbox and then navigate to each individual example (like `/dir1/dir2/example1`), so I don't see this as an issue.

Yes, same approach as in nebula.gl.

PR #332, opened 2020-10-09
Gatsby - Add header logo link

mletic:
Added the `HEADER_LINK_URL` field in the configuration that allows us to define the url that will be opened once the user clicks on the logo. By default it will take the user on the homepage (as before).

Also changed how links in the header work as Gatsby creates warnings if `Link` component is used for external links.
This also breaks the SPA experience if it isn't added properly.

One thing I had to expand is the GraphQL query in `ocular/modules/gatsby-theme-ocular/src/react/site-query.jsx`
If I don't add a new field there (`HEADER_LINK_URL`) this value won't be passed in the config that is available in the React components.

One thing that is confusing is the warning in the comment there that says:
`WARNING: DO NOT MODIFY THIS FILE MANUALLY IT WILL BE OVERWRITTEN.`

I can't see how this can be overwritten, or how to expand the available config unless I add this here.
So any help or clarification here is appreciated. 

@ibgreen @Pessimistress please review :)

### DIFF
--- a/docs/gatsby-theme-ocular/api-reference/options.md
+++ b/docs/gatsby-theme-ocular/api-reference/options.md
@@ -69,7 +69,7 @@ Each link entry has the following fields:
 | Field     | Type     | Name  |
 | ---       | ---      | ---   |
 | `name`    | `String` | Required. The displayed text in the header link. |
-| `href`    | `String` | Optional. The hyperlink you want to redirect to. This can be internal or external link. |
+| `href`    | `String` | Required. The hyperlink you want to redirect to. This can be internal or external link. |
 | `classnames` | `String` | Optional. The classname of the header link.
 | `index`      | `Int`    | Optional. Use 0 to push before first item. |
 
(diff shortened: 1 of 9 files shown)


Pessimistress (approved):
LGTM

ibgreen (approved):
LGTM I am not 100% on the details but I recall that I had a hunch we needed special processing for internal and external links, so adding that makes sense.

> One thing I had to expand is the GraphQL query in ocular/modules/gatsby-theme-ocular/src/react/site-query.jsx
If I don't add a new field there (HEADER_LINK_URL) this value won't be passed in the config that is available in the React components.

All fields should be declared in `source-nodes.js` to avoid graphQL issues. I think you are doing this.

> One thing that is confusing is the warning in the comment there that says:
WARNING: DO NOT MODIFY THIS FILE MANUALLY IT WILL BE OVERWRITTEN.

This may be an old warning. Previous versions of gatsby had an issue that this file wouldn't compile if it wasn't in a certain place, so the build copied it around. I think you can delete this comment / warning.

> you are using validate.js which doesn’t support boolean types, and I need them for IS_DOCS_INDEX or however we call it. I can do a string ‘true/false’ but that doesn’t make sense for me

I don't think we are tied to `validate JS` if there is a better solution. I also assume you could do additional programmatic validation after calling `validate`

PR #335, opened 2020-10-27
Support file extensions in Babel build

nrabinowitz:
At present, it isn't possible to specify a custom list of file extensions to build in `babel.config.js` - see https://github.com/babel/babel/issues/8652. There's an [open PR](https://github.com/babel/babel/pull/12216) for this support in Babel, but at the moment the only way to convince the Babel CLI to parse files with extensions other than `.js` and `.jsx` is to provide the list on the command line.

This PR adds support for custom extensions via `ocular build --extensions ".js,.ts"`, allowing the use of `ocular-dev-tools` for Typescript via `@babel/preset-typescript`.

### DIFF
--- a/modules/dev-tools/scripts/build.sh
+++ b/modules/dev-tools/scripts/build.sh
@@ -5,7 +5,7 @@ set -e
 DEV_TOOLS_DIR=`node -e "require('ocular-dev-tools/node/module-dir')()"`
 CONFIG=`node $DEV_TOOLS_DIR/node/get-config.js ".babel.configPath"`
 MODULES=`node $DEV_TOOLS_DIR/node/get-config.js ".modules" | sed -E "s/,/ /g"`
-EXTENSIONS=".es6,.js,.es,.jsx,.mjs"
+EXTENSIONS=`node $DEV_TOOLS_DIR/node/get-config.js ".babel.extensions"`
 
 check_target() {
   if [[ ! "$1" =~ ^es5|es6|esm ]]; then
@@ -47,9 +47,6 @@ build_monorepo() {
       case "$1" in
         --dist)
             TARGET=$2
-            shift ;;
-        --extensions)
-            EXTENSIONS=$2
             shift ;;
         *)
             echo -e "\033[91mUnknown option $1. ocular-build [--dist es5|es6|esm,...] [module1,...]\033[0m"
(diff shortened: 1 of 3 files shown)


modules/dev-tools/scripts/build.sh
```diff
@@ -5,6 +5,7 @@ set -e
 DEV_TOOLS_DIR=`node -e "require('ocular-dev-tools/node/module-dir')()"`
 CONFIG=`node $DEV_TOOLS_DIR/node/get-config.js ".babel.configPath"`
 MODULES=`node $DEV_TOOLS_DIR/node/get-config.js ".modules" | sed -E "s/,/ /g"`
+EXTENSIONS=".es6,.js,.es,.jsx,.mjs"
```
nrabinowitz, line 8: This is the current `@babel/cli` default - it was easier to specify here than leave out.

ibgreen (approved):
We typically handle things like this in ocular-dev-tools.js config file.

Regardless of how we do it, let's make sure we also update the docs.

In particular, I assume the babel defaults have to be repeated when overriding so maybe mention that in the docs and list them out?

I assume this is for adding typescript files - we could consider supporting those by default. i.e. the mere act of allowing typescript doesn't force repos to start using it...

nrabinowitz:
> We typically handle things like this in ocular-dev-tools.js config file.

Makes sense, I can move the settings there.

> Regardless of how we do it, let's make sure we also update the docs.

Will do.

> I assume this is for adding typescript files - we could consider supporting those by default. i.e. the mere act of allowing typescript doesn't force repos to start using it...

This is for TS via Babel, which seemed like the best option. I didn't add `.ts` to the list by default because then Babel tries to handle `.d.ts` files unless you explicitly ignore them, and fails. We could add this to the default Babel `ignore` setting, but it could still be a breaking change for any repo overriding `ignore`, so this seemed like the least intrusive option.

ibgreen:
> > We typically handle things like this in ocular-dev-tools.js config file.
> 
> Makes sense, I can move the settings there.
> 
> > Regardless of how we do it, let's make sure we also update the docs.
> 
> Will do.
> 
> > I assume this is for adding typescript files - we could consider supporting those by default. i.e. the mere act of allowing typescript doesn't force repos to start using it...
> 
> This is for TS via Babel, which seemed like the best option. I didn't add `.ts` to the list by default because then Babel tries to handle `.d.ts` files unless you explicitly ignore them, and fails. We could add this to the default Babel `ignore` setting, but it could still be a breaking change for any repo overriding `ignore`, so this seemed like the least intrusive option.

Understood. Seems reasonable as a first step.

Not clear why the `.d.ts` files should cause failures though. They are very valid TS syntax, big frameworks like THREE.js use this approach.

ibgreen:
I see, they are interpreted as `.ts`... how annoying, there must be a solution.

PR #338, opened 2020-11-29
Fix link mapping from README.md

Pessimistress:
For https://github.com/visgl/deck.gl/pull/5148


ibgreen (approved):
Maybe a good time to add at least one test case for the link mapping?

Pessimistress:
There is no point testing this component alone without the Gatsby node creation.

PR #342, opened 2020-12-25
Fix ToC node backfill and doc page creation export

ilijapuaca:
Fixes two issues that we ran into:
- In case where `Mdx` node gets created after table of contents (`DocsJson` node) is created, table of contents backfill did not work when `HOME_PATH` is specified. `tocNode` entries contain paths that don't have `HOME_PATH` prefix stripped (I assume it should remain that way), and it was being compared with relative path after the prefix has been removed
- `createDocPages` was not returning a promise, making `Promise.all` in `create-pages` resolve `docPromise` which is `undefined` immediately. This (sometimes) results in a warning message, noting that `createPage` is getting called inappropriately

On a side-note -- there is a comment in `on-create-node.js` noting that globals shouldn't be used to keep track of nodes in order to match them with ToC contents. Would it be possible to do node processing in `sourceNodes` instead of `onCreateNode`? Per docs, all the nodes will be present when `sourceNodes` is called, versus `onCreateNode` being called whenever each new node is created. That would allow for some cleanup as explicit handling for these race conditions would not be needed

### DIFF
--- a/modules/gatsby-theme-ocular/src/gatsby-node/create-pages/create-doc-pages.js
+++ b/modules/gatsby-theme-ocular/src/gatsby-node/create-pages/create-doc-pages.js
@@ -93,5 +93,5 @@ module.exports = function createDocPages({graphql, actions}, ocularOptions) {
     `Creating docs pages...`
   )();
 
-  createDocMarkdownPages({graphql, actions}, ocularOptions);
+  return createDocMarkdownPages({graphql, actions}, ocularOptions);
 };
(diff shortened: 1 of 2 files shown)


isaacbrodsky (approved):

ibgreen (approved):
This looks amazing. Seems like if we had some unit tests we could catch some of these things earlierl

> On a side-note -- there is a comment in on-create-node.js noting that globals shouldn't be used to keep track of nodes in order to match them with ToC contents. Would it be possible to do node processing in sourceNodes instead of onCreateNode? Per docs, all the nodes will be present when sourceNodes is called, versus onCreateNode being called whenever each new node is created. That would allow for some cleanup as explicit handling for these race conditions would not be needed

Yes - we should totally do a pass on this. If we can simplify this the ocular code base will be so much more accessible and easy to reason about.

PR #345, opened 2021-01-14
Fix: extensions list from config wasn't used when executing ESLint

ilyabo:


ibgreen:
This is good. One of the bigger problems with ocular-dev-tools is that the docs tend to get out of date. 

If every PR makes a small improvement we will soon be in good shape.

ibgreen:
Published as v0.2.1

PR #346, opened 2021-01-26
[WIP] Improved TypeScript support

ilyabo:


modules/dev-tools/templates/.eslintrc.js
```diff
@@ -0,0 +1,59 @@
+const typescriptConfigs = require('@typescript-eslint/eslint-plugin').configs;
+
+module.exports = {
+  parser: 'babel-eslint',
+  parserOptions: {
+    ecmaVersion: 2018
```
ibgreen, line 6:
```suggestion
    ecmaVersion: 2021
```

modules/dev-tools/templates/.eslintrc.js
```diff
@@ -0,0 +1,59 @@
...
+  rules: {
+    'guard-for-in': 0,
+    'no-inline-comments': 0,
+    'camelcase': 0
+  },
+  overrides: [
+    {
+      files: ['**/*.ts', '**/*.tsx'],
+      parser: '@typescript-eslint/parser',
+      env: {browser: true, es6: true, node: true},
```
ibgreen, line 22:
```suggestion
      env: {browser: true, es2021: true, node: true},
```
Why are we specifying the `envs` in this nested parserOptions and not up top?

modules/dev-tools/templates/.eslintrc.js
```diff
@@ -0,0 +1,59 @@
...
+    'prettier'
+  ],
+  rules: {
+    'guard-for-in': 0,
+    'no-inline-comments': 0,
+    'camelcase': 0
+  },
+  overrides: [
+    {
+      files: ['**/*.ts', '**/*.tsx'],
```
ibgreen, line 20: Some comments on these overrides, why we do need these would be really helpful. You have done a lot of research here, you could help future maintainers by explaining e.g why we need `sourceType: module`

modules/dev-tools/templates/.eslintrc.js
```diff
@@ -0,0 +1,59 @@
...
+      parser: '@typescript-eslint/parser',
+      env: {browser: true, es6: true, node: true},
+      parserOptions: {
+        sourceType: 'module',
+        project: './tsconfig.json'
+      },
+      plugins: ['@typescript-eslint'],
+      rules: {
+        ...typescriptConfigs['eslint-recommended'].rules,
```
ibgreen, line 29: Do you have any notes about which rules affect performance most. I still think performance is a bit of an issue with this plugin.

modules/dev-tools/templates/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
ibgreen, line 3:
```suggestion
    "noEmit": "true",
```

modules/dev-tools/templates/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    "forceConsistentCasingInFileNames": true,
+    "lib": [
+      "es2015",
```
ibgreen, line 16: Are the more recent available?

modules/dev-tools/package.json
```diff
@@ -41,12 +41,16 @@
     "@babel/core": "^7.8.7",
     "@babel/plugin-transform-runtime": "^7.8.7",
     "@babel/preset-env": "^7.8.7",
+    "@typescript-eslint/eslint-plugin": "^4.14.1",
```
ibgreen, line 44: We should bump the package.json version to `0.3.0`.

modules/dev-tools/templates/.eslintrc.js
ibgreen, line 20: I assume this also covers `d.ts` files?

ibgreen (approved):
Bump package.json to 0.3.0 to avoid automatic upgrades. 

Docs:
- Some kind of section about typescript support? 
- At least a bullet in what's new and upgrade-guide.md (I think renaming the eslint template my be a breaking change).
https://github.com/uber-web/ocular/tree/master/docs/dev-tools

modules/dev-tools/templates/.eslintrc.js
ilyabo: added a few

ilyabo: It does

modules/dev-tools/templates/.eslintrc.js
ilyabo: Not really. I'd guess `recommended-requiring-type-checking` are probably the most expensive.

modules/dev-tools/templates/.eslintrc.js
ilyabo: Yes, good point. Probably this can be pulled up

modules/dev-tools/package.json
ilyabo: Done. Shouldn't we add `-alpha.0`?
BTW, can we test it without publishing.

ibgreen:
Replaced by #361

PR #347, opened 2021-01-29
Support custom tag in publish script

Pessimistress:
### Background

react-map-gl is maintaining two parallel release tracks (mapbox v1/v2) and they should not be both tagged `latest`.

### Change List

Support custom npm tag in publish script


ibgreen (approved):
Nice

PR #348, opened 2021-02-17
[dev-tools] Fix coverage calculation

Pessimistress:


ibgreen (approved):

ibgreen:
I have some modules with code in `wip` directory would be good to exclude that too while we are at it.

Pessimistress:
This is just a template and is not by default used. You still need to create a .nycrc in each repo and import it. I think it's clearer to add any additional include/exclude there.

PR #349, opened 2021-02-23
[dev-tools] Change build targets

Pessimistress:
### This is a breaking change

- Drop support for IE and other obsolete browsers from ES5 and ESM targets - this saves about 20% off the transpiled bundle size
- Drop the ES6 dist target (less aggressive than the new ESM target) - we have been using this with the `esnext` field which is not a standard entry point
- Default babel config for unit tests now require Node.js >=10.24 - this is independent of the above changes and open for discussion. 10 is the smallest [maintained release](https://nodejs.org/en/about/releases/) and the first release that supports async functions.

### DIFF
--- a/docs/dev-tools/cli/ocular-build.md
+++ b/docs/dev-tools/cli/ocular-build.md
@@ -1,3 +1,7 @@
 # ocular-build
 
-Build the src
+Build the source.
+
+```bash
+ocular-build [module_name]
+```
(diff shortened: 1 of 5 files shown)


ibgreen (approved):
Node 14 is only for tests? 

It may not be well supported on all cloud environments. I think Google Cloud has finally added support but there may still be stragglers.

What will happen in repos where I have Node 12 test targets?

Pessimistress:
> What will happen in repos where I have Node 12 test targets?

We can certainly drop the target to node 12. The lower we go the more transpilation there will be.

PR #356, opened 2021-03-15
chore: Add ocular tests to ocular

ibgreen:
@ilijapuaca


ibgreen:
For some reason the CI doesn't complete, but at least we now run tests on CI...

PR #357, opened 2021-03-18
feat: ocular-tsify using ts-smoosh

ilijapuaca:
Integrates `ts-smoosh` work that @heshan0131 has done and exposes it through `ocular-tsify` command

### DIFF
--- a/modules/dev-tools/src/ts-smoosh/README.md
+++ b/modules/dev-tools/src/ts-smoosh/README.md
@@ -1,0 +1,116 @@
+## TS Smoosh
+
+Combine type decls with related source files.
+
+# Use
+
+Given a JavaScript file (or list of files) like so:
+
+```
+$ node ts-smoosh/bin ./src/some-file.js
+```
+
+Will produce `.ts` files using nearby `.d.ts` files.
+
+# Supported syntax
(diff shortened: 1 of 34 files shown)


modules/dev-tools/src/ts-smoosh/README.md
```diff
@@ -0,0 +1,116 @@
+## TS Smoosh
```
ibgreen, line 1: Move to docs?

modules/dev-tools/src/ts-smoosh/log.js
```diff
@@ -0,0 +1,41 @@
+const Colors = {
```
ibgreen, line 1: We are using probe.gl logging which has similar features, maybe leave a comment?

ibgreen (approved):

ibgreen:
@btford FYI your ts-smooch will live forever in ocular.

modules/dev-tools/src/ts-smoosh/README.md
ilijapuaca: Some of the implementation details like running tests aren't relevant outside the context of this source. Added a new doc `ocular-tsify.md` alongside other commands that contains a subset that is relevant to the end user

PR #359, opened 2021-03-21
chore: align gatsby module with dev-tools

ibgreen:
Move gatsby-theme-ocular back into modules tree.
Make ocular lint etc work on gatsby module


ibgreen:
@ilijapuaca

PR #365, opened 2021-06-19
For ES Modules: Look also for .ocularrc.cjs

Avnerus:
Hi! When working with a package that is defined as an ES module (`"type":"module"`, now [supported natively by Node 12 LTS](https://blog.sindresorhus.com/hello-modules-d1010b4e777b)), I get the following error:
```
Error: [ERR_REQUIRE_ESM]: Must use import to load ES Module: /****/.ocularrc.js
  require() of ES modules is not supported.
  require() of /***/node_modules/ocular-dev-tools/src/helpers/get-ocular-config.js is an ES module file 
as it is a .js file whose nearest parent package.json contains "type": "module" which defines all 
.js files in that package scope as ES modules.   Instead rename .ocularrc.js to end in .cjs, 
change the requiring code to use import(), or remove "type": "module" from /**/package.json.
 ```
I figured the easiest way to handle this is to add another search path option: `.ocularrc.cjs` (for example it is supported in [eslint](https://eslint.org/docs/user-guide/configuring/configuration-files)).  This PR fixed the problem for me, after I renamed `.ocularrc.js` to `ocularrc.cjs`.

### DIFF
--- a/docs/dev-tools/README.md
+++ b/docs/dev-tools/README.md
@@ -98,7 +98,7 @@ module.exports = env => {
 
 #### ocular-dev-tools.js
 
-A file `ocularrc.js` can be placed at the root of the package to customize the dev scripts. The config file may export a JSON object that contains the following keys, or a callback function that returns such object:
+A file `.ocularrc.js` (or `.ocularrc.cjs` for `type:module` projects) can be placed at the root of the package to customize the dev scripts. The config file may export a JSON object that contains the following keys, or a callback function that returns such object:
 
 - `lint`
   + `paths` (Arrray) - directories to include when linting. Default `['modules', 'src']`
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
It would have been nice with a comment before each case to explain what we are doing. 

I have not hit this problem in loaders.gl. I suppose we would need to set `type: module` in the top-level package.json to see it?

I like the possibility of moving ocular to `.mjs` files and use `import` / `export` internally instead of require.

Avnerus:
Thanks!
> It would have been nice with a comment before each case to explain what we are doing.
> 
Added some comments and also fixed the docs. I wasn't really sure about `ocular.config.js` and `ocular-dev-tools.config.js`. I saw them mentioned in the Gatsby theme documentation but not in the dev tools, so I marked them as _Backward compatibility_.
> I have not hit this problem in loaders.gl. I suppose we would need to set `type: module` in the top-level package.json to see it?
> 
Yes. I was forced to do this in my project to enable mocha along with typescript (using [this](https://github.com/TypeStrong/ts-node/issues/1007) solution) and Three.JS after they moved to pure ES6 modules. Might be a good idea to set this also in loaders.gl at some point.

PR #366, opened 2021-06-22
babel targets: explicit support for async functions in ESM, reduces runtime dependencies.

Avnerus:
Following the discussion on https://github.com/visgl/loaders.gl/pull/1532


ibgreen (approved):
Looks good, my only concern is that updating the frameworks to the latest ocular requires some work. I can start on loaders, math and luma.

Pessimistress (approved):

PR #367, opened 2021-06-22
chore: update browser list on `yarn bootstrap`

ibgreen:
- Make sure we are always seeing the same browser versions
- Unfortunately the database update is quite slow, adding 10-20 seconds to what is already a 14 minute CI run.
- I assume specifying actual browser versions would reduce the need for this step.


ibgreen:
@Avnerus

PR #368, opened 2021-06-22
chore(dev-tools): Fix Node 16 build

ibgreen:
Some weird issue with istanbul, as usual


ibgreen:
@Avnerus If you rebase your PR should build again.

Pessimistress:
Are you sure you should be checking in a 148k line file in `.yarn`?

ibgreen:
Hmm I tried using `yarn policies` to fix the version as I was having install issues. Didn't realize that it is storing the entire binary :-o. Good catch, I'll look into it.

PR #405, opened 2022-12-16
Update test harness to use ts-node and vite

Pessimistress:
- Node tests are run with `ts-node` instead of babel
- Browser tests are served with `vite`

Eventually, I would like to remove all babel dependencies and build with `tsc`. I am keeping them for now to avoid drastic changes in the consumer repos.


modules/dev-tools/package.json
```diff
@@ -89,13 +92,13 @@
...
-    "tape": "4.11.0",
+    "tape": "^4.11.0",
     "tape-promise": "^4.0.0",
-    "typescript": "^4.3.4",
-    "webpack": "^4.28.4",
-    "webpack-bundle-analyzer": "^3.0.3",
-    "webpack-cli": "^3.2.1",
-    "webpack-dev-server": "^3.1.14"
+    "typescript": "~4.6.0",
+    "ts-node": "~10.9.0",
```
Pessimistress, line 98: These two versions are the latest compatible combination that I can get to work. Restricting them may prevent us from using the latest TS features. I would expect `ts-node` to be not too far behind, given the big user base, so we can revisit this every now and then to upgrade.

PR #419, opened 2023-02-15
Disable print command

Pessimistress:


ibgreen (commented):
This is reverting a major feature change as there was big pushback from our ocular users due to a lack of transparency / "black box" concern: "what the heck is it doing".

Truth be told those sentiments seem difficult to completely remove. 

I assume there is a strong reason for doing this.

Pessimistress:
I understand that concern, but is printing the command itself the solution to it?

```
print_yellow "Running eslint in $DIRECTORIES..."
npx eslint --fix "$DIRECTORIES/**/*.$EXTENSIONS"
```

I would argue that the formatted echo message is much easier to understand.

If the feedback is that there is not enough structured output, then let's work on adding more human readable messages. If some user wants to see exactly what commands are being executed, then let's turn it on globally with a cli flag.

I read the output of Babel and Lerna closely during publish, and these arbitrary, complicated prints make it very hard to find the information I'm looking for.

ibgreen:
No, people already know it is linting, printing that in yellow text does not add anything meaningful. People who are having issues with their dev envs want to see what commands that are executed so they can understand what configs are being used, option set etc. Having to go into the ocular repo to figure this out is not popular.

Having the option to turn the logging on and off could be nice but I suspect that may not be trivial if commands call each other etc and possibly seems like overkill

But, ultimately ocular is an internal tools for vis.gl so we are not obliged to make others happy.

Issue #427, opened 2023-08-15
[configuration] using babel overrides throws error


chrisgervang:
https://github.com/uber-web/ocular/blame/69b7e75378ef249c3e723dc605dc6f7399439bd4/modules/dev-tools/src/configuration/get-babel-config.cjs#L111

I don't think this is functioning as intended. When I use `overrides` I get an `Error: babel.config.js: Configuration should be an exported JavaScript object.`

Shouldn't it be something like this?
```js
if (options.overrides) {
  config = deepMerge(config, {overrides: options.overrides});
}
```

PR #428, opened 2023-08-16
fix(dev-tools) babel config overrides nesting

chrisgervang:
For #427 

**Changes**
- Babel `overrides` should be nested under `overrides`
- The `overrides` type can be an array or undefined. Babel throws `ConfigError: .overrides must be an array, or undefined` otherwise.

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -2,7 +2,7 @@ const {getBabelConfig} = require('ocular-dev-tools/configuration');
 
 module.exports = getBabelConfig({
   react: true,
-  overrides: {
+  overrides: [{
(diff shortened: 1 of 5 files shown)


ibgreen (approved):
> ConfigError: .overrides must be an array, or undefined otherwise.

My only concern is that I wonder why has this has not been a problem before?
Is this error new in a recent version of babel?

chrisgervang:
Discussed with @Pessimistress, the api's intention is probably

```js
overrides: {
  // global overrides
  overrides: [{}] // babel's overrides api
}
```

PR #430, opened 2023-09-11
[dev-tools] Support building esm module with multiple entry points

Pessimistress:


modules/dev-tools/scripts/build.sh
```diff
@@ -26,13 +26,19 @@ build_src() {
...
 
-  if [ -e "src/index.ts" ]; then
-    ENTRY=src/index.ts
-  else
-    ENTRY=src/index.js
-  fi
+  # build cjs bundles
+  CJS_ENTRIES=`node $DEV_TOOLS_DIR/src/helpers/get-cjs-entry-points.js | sed -E "s/,/ /g"`
+
+  for P in ${CJS_ENTRIES}; do (
```
ibgreen, line 32: Nit: What is `P`?

modules/dev-tools/scripts/build.sh
```diff
@@ -26,13 +26,19 @@ build_src() {
...
+    if [ -e "src/${P}.ts" ]; then
+      ENTRY=src/${P}.ts
+    else
+      ENTRY=src/${P}.js
+    fi
 
-  esbuild $ENTRY --bundle --packages=external --format=cjs --target=es2015 --outfile=dist/index.cjs
+    echo "Bundling ${ENTRY}"
+    esbuild $ENTRY --bundle --packages=external --format=cjs --target=es2015 --outfile=dist/${P}.cjs
```
ibgreen, line 40: So we are building one bundle per entry? This means that if someone is using multiple entry points they will bundle multiple bundles with copies of the same code?

ibgreen (commented):
Nice addition, would be good to update docs and whats-new.

PR #439, opened 2023-11-03
feat(dev-tools): Inject __VERSION__ when building CJS dist

ibgreen:
- We get failures in loaders.gl 4.0 when using workers from applications that bundle via CJS. 
- We can see that __VERSION__ has not been defined in the CJS bundle.
- Add a define statement when building CJS, same as is done by all explicit esbuild commands in loaders.gl

### DIFF
--- a/modules/dev-tools/scripts/build.sh
+++ b/modules/dev-tools/scripts/build.sh
@@ -37,7 +37,7 @@ build_module_esm() {
     fi
 
     echo "Bundling ${ENTRY}"
-    esbuild $ENTRY --bundle --packages=external --format=cjs --target=node16 --outfile=dist/${P}.cjs
+    esbuild $ENTRY --bundle --packages=external --format=cjs --target=node16 --outfile=dist/${P}.cjs ----define:__VERSION__=\\\"$npm_package_version\\\"
   ); done
 }
 


ibgreen:
@igordykhta @belom88

Pessimistress:
This is just an arbitrary behavior. The right solution should be using the Babel config to build the cjs bundle (where loaders adds the version-inline plugin).

Another example of cjs entry point needing special build parameters is in deck.gl where jsx must be transformed for preact.

ibgreen:
> This is just an arbitrary behavior. 

Version handling doesn't seem that arbitrary to me, it seems like a generic thing. We are in the business of developing frameworks and having access to version numbers is a pretty basic thing. 

Whereas I agree that the preact example does indeed seem very arbitrary.

> The right solution should be using the Babel config to build the cjs bundle (where loaders adds the version-inline plugin).

Hmm, I was hoping that we were aggressively moving away from babel towards tools like esbuild and perhaps later even faster tools like bun. As long as we are tied to non-native tools like babel it seems to me that our build times are likely to remain long.

But those are just my opinions. What matters is that we need to fix this asap, so unless there is a way to add the parameter through .ocularrc or similar loaders.gl will need to build its cjs dists in a different way.

Pessimistress:
>  I was hoping that we were aggressively moving away from babel

The point is that if the source references something like `__VERSION__` then it should be handled consistently between the ESM build and the bundle, controlled by either babel config or esbuild config provided by the user. Adding a special case parameter to the bundle command opens the door for more discrepancy.

ibgreen:
> then it should be handled consistently between the ESM build and the bundle

Yes, that would be great. But do we have a mechanism to do something like that right now?

PR #446, opened 2024-01-29
[dev-tools] ESM module build command improvements

Pessimistress:
- Build the commonjs entry point and script bundle from the ESM entry point instead of source. This ensures that all versions of the library have the same transforms applied. It does create a build-time dependency (ESM must be rebuilt first) and may be inconvenient if debugging cjs or bundle locally.
- Babel can be disabled by adding `babel: false` to `ocularrc`. If so, TypeScript is expected to emit the compiled js files ( `typescript.project` in ocularrc)
- Add new export `ts-transform-version-inline` that replaces `babel-plugin-version-inline`
- Add new export `ts-transform-append-extension` to append `.js` extension to import statements (required by ESM)


modules/dev-tools/scripts/build.sh
```diff
@@ -20,25 +20,15 @@ build_src() {
   OUT_DIR=$1
   TARGET=$2
   check_target $TARGET
-  (set -x; BABEL_ENV=$TARGET npx babel src --config-file $CONFIG --out-dir $OUT_DIR --copy-files --source-maps --extensions $EXTENSIONS)
+
+  if [ ! -z "$CONFIG" ]; then
```
ibgreen, line 24: Why are we keeping babel? To make it easier to transition repos?

modules/dev-tools/src/configuration/get-esbuild-config.js
```diff
@@ -86,38 +66,51 @@ function umdWrapper(libName) {
...
+ * @param {String} opts.input - path to entry point
+ * @param {String} opts.output - output file path
+ */
+export async function getCJSExportConfig(opts) {
+  return {
+    entryPoints: [opts.input],
+    outfile: opts.output,
+    bundle: true,
+    format: 'cjs',
+    target: 'node16',
```
ibgreen, line 80: Node 16 is no longer supported, but I guess a grace period could make sense. Maybe add a comment above this line?

modules/dev-tools/src/helpers/get-config.js
```diff
@@ -18,7 +18,7 @@ configPath
   .split('.')
   .filter(Boolean)
   .forEach((path) => {
-    config = config[path];
+    config = config ? config[path] : undefined;
```
ibgreen, line 21: I am not quite sure I understand what this does. Maybe add a comment, why we split on `.`?

modules/dev-tools/ts-transform-append-extension/index.cjs
```diff
@@ -0,0 +1,51 @@
+/**
+ * Adapted from https://github.com/murolem/ts-transformer-append-js-extension
+ */
+const path = require('path');
+
+module.exports = function (program, pluginConfig, { ts }) {
```
ibgreen, line 6: Maybe name the function and add some JSDoc describing what this does and why we forked, to guide the maintainer?

modules/dev-tools/ts-transform-version-inline/index.cjs
```diff
@@ -0,0 +1,48 @@
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function (program, pluginConfig, { ts }) {
```
ibgreen, line 4: Ditto.

Can these go into a ts-plugins directory? Is it necessary to have them in the root?

modules/dev-tools/ts-transform-append-extension/index.cjs
ibgreen: I assume ESM plugins are not yet supported...

ibgreen (approved):
Looks promising. Adding some docs on function etc would be neat.

modules/dev-tools/ts-transform-append-extension/index.cjs
Pessimistress: ESM support is experimental, and requires additional dependencies

Pessimistress: Moved to .ts and added a build step

PR #455, opened 2024-02-22
feat(dev-tools): add ts-transform-remove-glsl-comments

Pessimistress:
TypeScript transform to remove comments and unnecessary white space from GLSL source.

A template string is processed as GLSL source if:
   1. the file matches the pattern specified in the plugin config; or
   2. it is tagged as `` glsl`...` ``

Usage with ts-patch:
```
   {
      "plugins": [
        {
          "transform": "ocular-dev-tools/ts-transform-remove-glsl-comments",
          "pattern": ["*.glsl.ts"]
        }
      ]
   }
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 dist/
+test-case-*.ts
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 **/dist*/**/*.js
+test-case-*.ts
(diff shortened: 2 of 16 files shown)


Pessimistress:
@ibgreen since template tag (``glsl`...` ``) only has cosmetic value would you like to remove it when transpiling with this plugin?

ibgreen:
> since template tag (glsl`...` ) only has cosmetic value would you like to remove it when transpiling with this plugin?

This is used to get syntax highlighting in embedded shaders which is awesome. I am not excited about removing it from the source. It is fine to strip it during the compilation of course.

PR #456, opened 2024-02-29
Add ts-transform-inline-webgl-constants

Pessimistress:
Port of https://github.com/visgl/luma.gl/tree/master/dev-modules/babel-plugin-inline-webgl-constants


ibgreen:
Nice. Can we remove these from luma.gl now?

Also should check luma.gl bundle sizes before and after this, should make a nice improvement!

deck.gl and apps shouldn't be using constants anymore once the v9 API port is complete, so I expect this will mostly be to keep luma.gl size down.

Pessimistress:
> Can we remove these from luma.gl now?

Yes? The only place still using the Babel plugins are the websites. They can switch to build from dist. However I can see the Babel plugins being useful to a user of our libraries.

PR #458, opened 2024-03-07
feat(devtools): Add bootstrap nobuild option

ibgreen:
Hoping to achieve a clean run with only a single install. A matter of principle!


Pessimistress (approved):

modules/dev-tools/scripts/bootstrap.sh
```diff
@@ -32,5 +32,7 @@ else
   yarn link $packageName
 fi
 
-# build the submodules
-npm run build
+# build the submodules unless `nobuild` argment is passed
+if [[ ! "$1" = "nobuild" ]]; then
```
Pessimistress, line 36: - IMO `skip-build` would be clearer
- Update docs?

PR #460, opened 2024-04-11
feat(dev-tools): Inline ESM-only dependency

donmccurdy:
For discussion. 

As of v9, deck.gl provides both ESM and CJS entrypoints. However, some dependencies — [notably `@mapbox/tiny-sdf`](https://github.com/mapbox/tiny-sdf/issues/53) — are ESM-only and do not include CJS entrypoints. This seems to be a problem in certain environments as noted in https://github.com/visgl/deck.gl/issues/7735, which may require CJS versions of dependencies.

This PR implements a possible solution in which the `@mapbox/tiny-sdf` dependency is inlined into the compiled CJS bundle. Optionally, we could do the same for the ESM bundle, and move the package from dependencies to devDependencies. The module is small (1.2kB), and can be tree-shaken, so the inlining seems acceptable.

I assume that if we decide to take this approach, we'd change the API such that the package name is not hard coded into ocular-dev-tools, and we instead either pass a list of package names in the project's ocular config, or attempt to automatically detect which dependencies do not offer CJS entrypoints.

Alternatives considered:

- Replace `@mapbox/tiny-sdf` with an alternative
- Fork `@mapbox/tiny-sdf`, add CJS build
- Drop CJS support in deck.gl

### DIFF
--- a/modules/dev-tools/src/configuration/get-esbuild-config.ts
+++ b/modules/dev-tools/src/configuration/get-esbuild-config.ts
@@ -4,7 +4,7 @@ import {join} from 'path';
 import util from 'util';
 import {getOcularConfig} from '../helpers/get-ocular-config.js';
 import ext from 'esbuild-plugin-external-global';
-import type {BuildOptions} from 'esbuild';
+import type {BuildOptions, Plugin} from 'esbuild';
 
 /**
  * Get list of dependencies to exclude using esbuild-plugin-external-global
@@ -67,6 +67,26 @@ function umdWrapper(libName: string | undefined) {
   };
 }
 
+/**
+ * ESBuild plugin to inline important ESM-only dependencies, for CJS compatibility.
+ * Reference: https://github.com/evanw/esbuild/issues/3442
+ */
+const inlineESMOnly = (): Plugin => {
+  const packageRoot = process.cwd();
+  const root = join(packageRoot, '../..');
+
+  return {
+    name: 'inline-esm-only',
+    setup(build) {
+      // TODO: Detect ESM-only from package.json, instead of hard-coding package names?
+      build.onResolve({filter: /^@mapbox\/tiny\-sdf$/}, () => {
+        const path = join(root, 'node_modules/@mapbox/tiny-sdf/index.js');
+        return {path, external: false};
+      });
+    }
+  };
+};
+
 /** Returns esbuild config for building .cjs bundles */
 export async function getCJSExportConfig(opts: {
   input: string;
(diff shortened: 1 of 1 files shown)


ibgreen:
This may be a useful capability to have in ocular. 

That said, my general view is that we should be very selective with our use of dependencies, and for small things, we can just fork the code if the original publishers are no longer active.

In this case it looks like one 140-line code file that has only had one or two micro fixes in the last two years, so why not just copy it into our repo, instead of fighting the JS build systems.

https://github.com/mapbox/tiny-sdf/blob/main/index.js

donmccurdy:
I'm open to that! It is indeed a small package. My experience over the past 1-2 years has been more and more packages going ESM-only intentionally, so I don't think the issue is specific to unmaintained dependencies, but no matter.

ibgreen:
> My experience over the past 1-2 years has been more and more packages going ESM-only intentionally, so I don't think the issue is specific to unmaintained dependencies, but no matter.

That is a good point. A question then is how long should deck.gl be trying to stem that tide? 

> I use Microbundle for building most of my personal repositories

Yes, in JS there is always another tool... I spent a lot of time learning webpack and now esbuild, so from that perspective I am slightly wary of making another switch so soon, and like the idea of staying with major tools (maybe microbundle is bigger than I think).

donmccurdy:
> That is a good point. A question then is how long should deck.gl be trying to stem that tide?

I suppose we'll know when we know. 😆 

> ...like the idea of staying with major tools (maybe microbundle is bigger than I think).

Not particularly, I'd say it's a highly opinionated wrapper around Rollup, in the same way that Ocular is (now) a highly opinionated wrapper around ESBuild. Definitely not suggesting a switch in tools, Microbundle is probably _too_ opinionated for all of VisGL ... just helpful to compare default behaviors!

***

Suggestion: Let's wait to hear if TinySDF remains problematic in v9. I suspect it will, but no point taking action until we're sure. Then we can either fork it or inline it.

modules/dev-tools/src/configuration/get-esbuild-config.ts
```diff
@@ -67,6 +67,25 @@ function umdWrapper(libName: string | undefined) {
...
+ * ESBuild plugin to inline important ESM-only dependencies, for CJS compatibility.
+ * Reference: https://github.com/evanw/esbuild/issues/3442
+ */
+const inlineESMOnly = (): Plugin => {
+  const packageRoot = process.cwd();
+
+  return {
+    name: 'inline-esm-only',
+    setup(build) {
+      // TODO: Detect ESM-only from package.json, instead of hard-coding package names?
```
Pessimistress, line 80: If this is difficult to do, can we make it a ocularrc config instead of hard-coding it here?

PR #461, opened 2024-04-23
feat(dev-tools): add new modes to ocular-publish

Pessimistress:
- Upgrade lerna to the latest
- New publish modes to support tag-locally, publish-on-CI scenario

cc @felixpalmer @donmccurdy

### DIFF
--- a/modules/dev-tools/docs/cli/ocular-publish.md
+++ b/modules/dev-tools/docs/cli/ocular-publish.md
@@ -18,8 +18,8 @@ npx ocular-publish [mode] [npm-tag]
 
 - `beta` - bump pre-release version and publish with beta flag.
 - `prod` - bump patch version and publish.
-- `version-beta` - bump pre-release version only.
-- `version-prod` - bump patch version only.
+- `version-only-beta` - bump pre-release version only.
+- `version-only-prod` - bump patch version only.
 - `from-git`: publish from the current git tag.
 
 ## npm-tag
--- a/modules/dev-tools/docs/cli/ocular-publish.md
+++ b/modules/dev-tools/docs/cli/ocular-publish.md
@@ -18,6 +18,9 @@ npx ocular-publish [mode] [npm-tag]
 
 - `beta` - bump pre-release version and publish with beta flag.
 - `prod` - bump patch version and publish.
+- `version-beta` - bump pre-release version only.
+- `version-prod` - bump patch version only.
+- `from-git`: publish from the current git tag.
 
 ## npm-tag
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules/
 dist/
 
 .idea/
+.nx/
 yarn-error.log
 .DS_Store
 .vscode/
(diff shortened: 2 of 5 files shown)


modules/dev-tools/docs/cli/ocular-publish.md
```diff
@@ -18,6 +18,9 @@ npx ocular-publish [mode] [npm-tag]
 
 - `beta` - bump pre-release version and publish with beta flag.
 - `prod` - bump patch version and publish.
+- `version-beta` - bump pre-release version only.
```
ibgreen, line 21: These docs looks quite terse, perhaps we can be a bit more verbose. Without reading the source I can't quite tell what these do.

Nit: The docs make it look like these two modules don't actually publish? If that is true, should we call them ocular-bump instead of ocular-publish then? or maybe `ocular-publish bump-beta` or `ocular-publish bump beta` `ocular-publish bump-only beta`

modules/dev-tools/docs/cli/ocular-publish.md
```diff
@@ -18,6 +18,9 @@ npx ocular-publish [mode] [npm-tag]
...
+- `version-prod` - bump patch version only.
+- `from-git`: publish from the current git tag.
```
ibgreen, line 23: Wording nit: What is "publish from" intended to convey? I assume we always publish from the checked out commit. So, what does this actually do? Scrapes git to see if there are tags on the currently checked out commit, and if so picks one of those? 

```suggestion
- `from-git`: publish with the npm tag set to the git tag of the currently checked out commit. Fails if the checked out commit doesn't have a tag.
```

ibgreen (approved):
Nice to see latest lerna. Docs could use a little TLC so less context is needed.

modules/dev-tools/docs/cli/ocular-publish.md
donmccurdy: Related:

https://lerna.js.org/docs/features/version-and-publish#from-package

Lerna has both `from-git` and `from-package` commands. I'm not sure why `from-git` doesnt't seem to be documented. For our purposes, probably either would be fine.

modules/dev-tools/docs/cli/ocular-publish.md
donmccurdy: In npm, yarn, and lerna, the "version" verb means incrementing the release number, and the "publish" verb means pushing that change to a registry. A postversion script might also run publish if projects prefer to do it all in one step, which seems common. 

I'd prefer if ocular could reuse these terms as much as possible — "bump" has no precise meaning to me. 

I do see your point that `ocular-publish version-beta` feels unclear, though...

Pessimistress: Can't use two words as the second argument is currently used as a custom NPM dist tag. This can of course change because it's not actively used by any repo and it seems like we are going with `publishConfig.tag` instead (I think that's a better approach).

Otherwise, does `version-only-beta` sound better?

PR #465, opened 2024-04-28
fix(dev-tools): drop deprecated node option

Pessimistress:
- Drop deprecated node option `--loader`, use `--import`
- Requires Node 18 LTS+

### DIFF
--- a/modules/dev-tools/src/helpers/esm-alias.ts
+++ b/modules/dev-tools/src/helpers/esm-alias.ts
@@ -6,16 +6,29 @@
 import path from 'path';
 import fs from 'fs';
 import {pathToFileURL} from 'url';
-import {resolve as resolveTs, getFormat, transformSource, load} from 'ts-node/esm';
 import {getValidPath, ocularRoot} from '../utils/utils.js';
-export {getFormat, transformSource, load};
+
+type ResolveHook = (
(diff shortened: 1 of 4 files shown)


modules/dev-tools/src/helpers/esm-alias.ts
```diff
@@ -6,16 +6,29 @@
 import path from 'path';
 import fs from 'fs';
 import {pathToFileURL} from 'url';
-import {resolve as resolveTs, getFormat, transformSource, load} from 'ts-node/esm';
 import {getValidPath, ocularRoot} from '../utils/utils.js';
-export {getFormat, transformSource, load};
+
+type ResolveHook = (
```
ibgreen, line 11: Nit: Add TSDoc on the type (and fields if possible). Specify if this is this an ocular type or are we compensating for esbuild 's lack of types, if so which version of esbuild was this written for?

ibgreen (approved):
LGTM. More comments usually better in this type of code.

modules/dev-tools/src/helpers/esm-alias.ts
Pessimistress: This is Node.js API. I'll add a comment.

PR #470, opened 2024-08-22
fix: test node-debug

belom88:
This PR fixes `yarn test node-debug`. I've tested it manually in @loaders.gl repository.

### DIFF
--- a/modules/dev-tools/src/test.ts
+++ b/modules/dev-tools/src/test.ts
@@ -44,7 +44,7 @@ switch (mode) {
     break;
 
   case 'node-debug':
-    runNodeTest(resolveNodeEntry('test'), '', true);
+    runNodeTest(resolveNodeEntry('test'), '', {breakAndInspect: true});
     break;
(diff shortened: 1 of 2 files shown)


modules/dev-tools/src/test.ts
```diff
@@ -43,6 +43,10 @@ switch (mode) {
     }
     break;
 
+  case 'node-debug':
+    runNodeTest(resolveNodeEntry('test'), '', true);
```
ibgreen, line 47: boolean parameters are "evil", no way for reader to know what this means. Use a named option instead?
```suggestion
    runNodeTest(resolveNodeEntry('test'), '', {breakAndInspect: true);
```

ibgreen (approved):

Pessimistress (approved):

modules/dev-tools/src/test.ts
belom88: Done

